PR #14, opened 2014-09-06
Allow setting CHARACTER SET for the database

pstiasny:
Adds the `MYSQL_CHARSET` environmental variable to go along with MYSQL_DATABASE.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -38,7 +38,7 @@ if [ "$1" = 'mysqld' ]; then
 		EOSQL
 		
 		if [ "$MYSQL_DATABASE" ]; then
-			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ;" >> "$tempSqlFile"
+			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ${MYSQL_CHARSET:+CHARACTER SET $MYSQL_CHARSET} ;" >> "$tempSqlFile"
 		fi
 		
 		if [ "$MYSQL_USER" -a "$MYSQL_PASSWORD" ]; then
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -38,7 +38,7 @@ if [ "$1" = 'mysqld' ]; then
 		EOSQL
 		
 		if [ "$MYSQL_DATABASE" ]; then
-			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ;" >> "$tempSqlFile"
+			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ${MYSQL_CHARSET:+CHARACTER SET $MYSQL_CHARSET} ;" >> "$tempSqlFile"
 		fi
 		
 		if [ "$MYSQL_USER" -a "$MYSQL_PASSWORD" ]; then
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -38,7 +38,7 @@ if [ "$1" = 'mysqld' ]; then
 		EOSQL
 		
 		if [ "$MYSQL_DATABASE" ]; then
-			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ;" >> "$tempSqlFile"
+			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ${MYSQL_CHARSET:+CHARACTER SET $MYSQL_CHARSET} ;" >> "$tempSqlFile"
 		fi
 		
 		if [ "$MYSQL_USER" -a "$MYSQL_PASSWORD" ]; then


damonkreft:
+1 too

tianon:
The reason I'd rather not use UTF-8 as the default is that upstream doesn't -- as much as I personally think UTF-8 is a more sane default than Latin1, upstream doesn't agree, and I'd rather have us be closer to upstream than not.  Also, other drivers will still default to `latin1` (see http://archive.oreilly.com/pub/post/turning_mysql_data_in_latin1_t.html for an example of one person's woes in this regard with PHP's driver), so changing the default away from that really ought to be a conscious user choice.

So, I'm currently leaning towards this being a documentation issue.

Something like this should do the trick:
```Dockerfile
FROM mysql:5.6
RUN { \
    echo '[mysqld]'; \
    echo 'character-set-server = utf8'; \
    echo 'default-collation = utf8_unicode_ci'; \
} > /etc/mysql/conf.d/charset.cnf
```

Combine that with automated builds (https://docs.docker.com/docker-hub/builds/) and repository links (https://docs.docker.com/docker-hub/builds/#repository-links) and it's reasonably easy to have an up-to-date image built `FROM` this one with a modified default character set.

gaomd:
@tianon Your solution to default UTF-8 charset configuration is neat.

But both this pull request and your simpler script just provide an environmental variable, which I think is way better than configuration file solution, I think all we need is a chance to set charset upon container creation.

shuoli84:
It is a really neat docker image, any reason why this change not merged? This is the only issue I am having right now.

thaJeztah:
@shuoli84 looking at the discussion above, it looks like the pull-request hasn't been updated with the suggestions made in https://github.com/docker-library/mysql/pull/14#issuecomment-58573069

Also (https://github.com/docker-library/mysql/pull/14#issuecomment-73567710) the image should match the default settings used "upstream" (in MySQL) itself and the maintainers lean to "not overloading the image with environment variables".

rysiekpl:
This is also very relevant to my interests and doesn't seem like a huge and potentially destructive change. Is there a chance it will get merged?

md5:
@rysiekpl I don't see this getting merged unless someone makes the updates @tianon suggested in https://github.com/docker-library/mysql/pull/14#issuecomment-58573069

Either @pstiasny will need to update this branch to make those changes or someone who has time to test them will need to submit a new PR.

pstiasny:
All right, I've reapplied @tianon's simplified patch.

jaredm4:
Anyone have a few minutes to test the patch?

md5:
@pstiasny :metal:

yosifkit:
LGTM, can we get it applied to 5.5 as well :heart_eyes:?  Then we need to make sure to document it (probably after https://github.com/docker-library/docs/pull/191 merges).

```console
$docker run -d -p 3306:3306 -e MYSQL_ROOT_PASSWORD=mysecretpassword -e MYSQL_C
HARSET=utf8 -e MYSQL_DATABASE=lol mysql:5.7

# connect to db
mysql> show variables like "character_set_database";
+------------------------+-------+
| Variable_name          | Value |
+------------------------+-------+
| character_set_database | utf8  |
+------------------------+-------+
1 row in set (0.00 sec)
```

rysiekpl:
@tianon I believe the second variable should be `collation-server` instead of `default-collation`; [with the latter an "unknown variable" error is thrown in MySQL 5.7](http://stackoverflow.com/questions/16339922/change-default-collation-character-set).

md5:
It would be cool to see some tests go along with this. Something simple like the `show variables` check that @yosifkit put in his comment would be great.

andrewserff:
I'm new to docker, but I ran into this yesterday, so I'm glad you guys were working on it!  I just tested it out locally and it seems to be working for me from both docker and docker-compose.  Thanks!

pstiasny:
I've finally had some time and updated the patch to also work for 5.5 as @yosifkit requested.

thaJeztah:
Great, @pstiasny! I think this also requires a change to the documentation https://github.com/docker-library/docs/tree/master/mysql#environment-variables. Do you have a chance to create a PR for that as well?

yosifkit:
LGTM

tianon:
LGTM

buckett:
I've created a documentation update for this.

thaJeztah:
Great @buckett!

ping @tianon @yosifkit this already has two LGTMs so looks like this is ready to merge?

see https://github.com/docker-library/docs/issues/237 for the associated docs changes.

yosifkit:
Not totally sure why this was not merged.  But I am now slightly hesitant, since this would only affect the database created by `$MYSQL_DATABASE` and not any they may create later.  Is that a reasonable limitation?

Maybe we should instead make a conf file to set the default charset for all databases when they pass in the env.

thaJeztah:
@yosifkit looks like there's an almost ready-to-use example to implement that in https://github.com/docker-library/mysql/pull/14#issuecomment-73567710

tianon:
I've just tested, and the following shorter/simpler example works, too:
```
[mysqld]
character-set-server = utf8
```

thaJeztah:
@tianon I think the default collation should also be set; https://dev.mysql.com/doc/refman/5.7/en/charset-applications.html

Also, I think it would make sense to configure the client accordingly: https://dev.mysql.com/doc/refman/5.7/en/charset-configuration.html

buckett:
The reason it's nice to have it available as an environmental variable is that it's something lots of services like (a UTF-8 database) and it makes it much easier to create a MySQL container without having to copy a file into the correct place. For example if you can just start a container with:
```
docker run -p 3306:3306 -e MYSQL_DATABASE=sakai -e MYSQL_CHARSET=utf8 mysql
```

yosifkit:
Yes, +1 to having an environment variable. I would like the variable to create a file in `/etc/mysql/conf.d/` to set it for the whole server rather than just the database created with `MYSQL_DATABASE`. We should probably also have a `MYSQL_COLLATION` to go along with it.

mozillazg:
+1 for `MYSQL_CHARSET` and `MYSQL_COLLATION`.

gotche:
+1

buckett:
If we end up just setting the character set for the newly created DB the it would be clearer to use the environmental property `MYSQL_DATABASE_CHARSET` to make it clearer that it's just for the new DB and then we could use `MYSQL_CHARSET` to change the default for all DBs.

Trying to do this for all DBs could become confusing if people are also placing their own files in `/etc/mysql/conf.d/` and aren't expecting there to be anything there already.

kung-fu-tzu:
For those poor souls googling for the *dockerized* solution for MySQL 5.6, create a `./mysql-charset/Dockerfile` with:
```dockerfile
FROM mysql:5.6

RUN { \
      echo '[mysqld]'; \
      echo 'character-set-server=utf8'; \
      echo 'collation-server=utf8_general_ci'; \
      echo '[client]'; \
      echo 'default-character-set=utf8'; \
    } > /etc/mysql/conf.d/charset.cnf
```
then:
```bash
docker build -t mysql-charset ./mysql-charset/
```
and then use the `mysql-charset` image instead of `mysql:5.6`.

Check the charset/collation with the following query:
```sql
SHOW VARIABLES LIKE 'char%'; SHOW VARIABLES LIKE 'collation%';
```

Do not forget reading [the original comment above](https://github.com/docker-library/mysql/pull/14#issuecomment-73567710) on this kind of solution.

riquito:
For the greater good, PLEASE, do not use the value "utf8" as default, it's an invented encoding whose length is just 3 bytes instead of 4. Use "utf8mb4" instead. MySQL's "utf8" supports just the Basic Multilingual Plane and doesn't support the Astral Planes. If you ever need to deal with oriental languages or emoticons you're in for a lot of pain using the fake utf8.

mksm:
For now I just created the database manually using the default charset I want.

+1 for the ENV variable.

carolinux:
Please merge this

gtklocker:
:+1:

md5:
@carolinux This PR isn't ready to be merged (cf. https://github.com/docker-library/mysql/pull/14#issuecomment-103651719)

It looks like the holdup is the question of whether to have `MYSQL_CHARSET` affect all database created in the container or just the default database, along with the possibility of needing to specify the collation as well.

carolinux:
Indeed it seems more reasonable to have the setting affect all dbs. My 2c is that it would be counter intuitive the other way around.

heschmidt04:
I advocate for a trifecta approach. Have the client reflect the MySQL daemon settings. 
Being able to set both MySQL charset and collation would indeed be valuable to set at the database. 
There is more to the complete setup and making sure all linked Docker instances have all consistent settings at each level. 

Levels
	a. MySQL  database creation with charset and collation with Docker vars MYSQL_CHARSET and MYSQL_COLLATION 
	b. MySQL  runtime  ([client], [mysql], [mysqld]) for the MySQL startup config file my.cnf to enforce global variables 
	c. MySQL client applications - making sure the OS shell environment has the right setting for LANG or it will invalidate the work of a and b. 

Going forward could the passing of the variables MYSQL_CHARSET and MYSQL_COLLATION do these functions? 

	1. Handle the OS level LANG setting for the docker image and make the end user aware to set that for any linked images that would use mysql client. If LANG isn't set; then session variables default to latin1. 
	LANG example:
	'''LANG=en_US.UTF-8'''
	
	2. Somehow get the settings into the image where it can be viewed. 
	And/or  update/create in file docker-compose.yml with the information for the path to the local config directory inside the code base with a standard volume name. 
	Have the process create the mock up of the volume my.cnf with correct dependent settings passed by charset and collation so that all of these MySQL parameters are kept in sync
	
	Charset and Collation parameters to keep in sync at global and session level: 
		'''character_set_client, character_set_connection, character_set_database, character_set_filesystem, character_set_results, character_set_server, character_set_system, collation_connection, collation_database, collation_server'''
	
           Docker-compose.yml volume example: 
                  '''volumes:
                          -  ./path_to_code_config/mysql-config-dir/:/etc/mysql/conf.d:ro'''

	Docker volume my.cnf example: 
	'''[client]
	default-character-set = utf8
	[mysql]
	
	[mysqld]
	character-set-filesystem = utf8
	character-set-client = utf8
	character-set-server = utf8
	collation-server = utf8_unicode_ci'''

ye:
+1 allowing charset and collation setting in a configuration file https://github.com/docker-library/mysql/pull/14#issuecomment-73567710

As far as the default mysql configuration, I __strongly recommend use `utf8mb4` charset instead of `utf8`__. 

The reason being that MySQL has been notorious about choosing the wrong default encoding (latin1) and then chose a broken UTF-8 implementation (3-byte vs. the full 4-byte Unicode standard, details here [1]) and caused many gotchas and issues (see [2]).

[1]: https://mathiasbynens.be/notes/mysql-utf8mb4
[2]: http://stackoverflow.com/search?q=mysql+charset

heschmidt04:
Hi Ye,

From what I found in my testing the utf8mb4 breaks the mysql2 gem and the
database objects are not created. I also get a non supported error message.
I’d need to do more to gems and use a work around for now even though it is
supposed to be resolved in 0.3.12
MySQL gem version we are using for mysql2 is 0.3.17
You can read up more on this thread in
https://github.com/brianmario/mysql2/issues/367 for it.

ye:
@heschmidt04 `utf8mb4` is a MySQL invention and is only recognizable in MySQL(see [1]), in everywhere else, it is called `UTF-8` or `utf8`. The initial implementation of unicode in MySQL uses 3-BYTE UTF-8, which only covers plane-0, aka the Basic Multilingual Plane, 1/16 of all unicode code points, or about 6.25% of all unicode characters. (see [2]) 

While I am exactly sure why `utfbm4` didn't work from your mysql gem build, it's definitely the way to go if you are manually setting it to ensure full coverage of unicode characters.

When `utf8mb4` charset is used, emoji is working.
screen shot 2015-08-25 at 6 39 17 pm


And if `utf8` charset is used, emoji is not working.
screen shot 2015-08-25 at 6 39 51 pm


[1]: https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html
[2]: https://en.wikipedia.org/wiki/Plane_(Unicode)](https://en.wikipedia.org/wiki/Plane_(Unicode)

nemonweb:
+1 please add

takashi:
+1 please

ltangvald:
I'm a bit hesitant about adding environment variables for this. Simply because there are so many server options (many of which have defaults that may no longer be that sensible) and we can't add environment variables for them all.
The image also has a feature that needs to be documented better, in that you can pass any server option in by appending it to the run command. Wouldn't «docker run [..] -d mysql --character-set-server=cs» and similar work fairly well for this?

rysiekpl:
> Wouldn't «docker run [..] -d mysql --character-set-server=cs» and similar work fairly well for this?

No, because you want to be able to do this in `docker-compose`, for example, without mucking with the `command` option.

yosifkit:
Setting command in the compose file is exactly correct; nothing special or extra needed.  Like @ltangvald noted, the entrypoint in the image has an undocumented feature; when the command starts with `-` then it assumes that those are flags to `mysqld`.

```yaml
  command: --character-set-server=cs

# not this:
  command:
    - "--flag-one=true --flag-two"

# this is ok:
  command:
    - --flag-one=true
    - --flag-two
```

OmarIthawi:
I've read the comments here and I understood that there are many requirements here. However I in my case I needed only to customize the default database. I'd *claim* that most of the users would need only this requirement.

The thing is, it was very handy to just pass few variables to using docker-compose and an instance of this image would have the database ready. I was only unhappy because I liked a little bit more control over the database's collate (I mainly develop bilingual apps). I see that the original contribution, and PR #120 are an answer for this. For example it helped @Zophar78.

heschmidt04:
Agreed with @OmarIthawi -- if the charset and collation can be chosen at the command line at least.
I'll have to try @yosifkit idea of passing --variables and see if they are applied to my.cnf. Very cool if it works for me. My concern is that the undocumented feature would change. 
@ye great points on mojibake

MVP is to get the charset and collation at the cmd line prompt. 
I can make sure the my.cnf and the LANG env variable can be set to keep them in sync. 
I was making folks aware that even if they set their database correctly they might see in the mysql status command that the session vars wouldn't match up to global settings.  Info to troubleshoot and ideas of what needs to be kept in sync. 

For object creation - as long as the global vars and the database are in sync the objects will follow. The client session vars are more for how the data comes back to the end user. To avoid mojibake

For parsing by docker, It does mean that incorrect combinations for charset and collation may cause weird outlier failures. Then we get to search them out :) 

Thanks,
Heidi

drobune:
:+1: please

arthurbarros:
+1

TakesxiSximada:
+1

ptahchiev:
+1

gochev:
Who in 2016 is not using UTF8 ? And to have no possibility to set UTF-8 is a joke.
+1

adius:
Here a simple docker-compose file to get started, with a unicode enabled version.
Add more services to make it useful (e.g. php+apache or phpmyadmin).
SQL (and `.sh`) files in the `sql-scipts` directory will be executed during startup.

```yaml
version: '2'
services:
  mysql:
    image: mysql:5.7
    ports:
      - '3306:3306'
    volumes:
      - ./sql-scripts/:/docker-entrypoint-initdb.d/
    command:
      - --character-set-server=utf8mb4
      - --collation-server=utf8mb4_unicode_ci
      - --skip-character-set-client-handshake
    environment:
      MYSQL_ROOT_PASSWORD: pass
      MYSQL_DATABASE: test
```

tracycool:
@peter-leonov nice

PR #18, opened 2014-10-21
Allow creating multiple databases on startup.

tonio-ramirez:
This can be useful in some cases.  For example, if you're auditing entities, and want to store the audit information on a separate database.

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -22,14 +22,18 @@ if [ -z "$(ls -A /var/lib/mysql)" -a "${1%_safe}" = 'mysqld' ]; then
 	EOSQL
 	
 	if [ "$MYSQL_DATABASE" ]; then
-		echo "CREATE DATABASE IF NOT EXISTS $MYSQL_DATABASE ;" >> "$TEMP_FILE"
+		for DATABASE in $MYSQL_DATABASE; do
+			echo "CREATE DATABASE IF NOT EXISTS $DATABASE ;" >> "$TEMP_FILE"
+		done
 	fi
 	
 	if [ "$MYSQL_USER" -a "$MYSQL_PASSWORD" ]; then
 		echo "CREATE USER '$MYSQL_USER'@'%' IDENTIFIED BY '$MYSQL_PASSWORD' ;" >> "$TEMP_FILE"
 		
 		if [ "$MYSQL_DATABASE" ]; then
-			echo "GRANT ALL ON $MYSQL_DATABASE.* TO '$MYSQL_USER'@'%' ;" >> "$TEMP_FILE"
+			for DATABASE in $MYSQL_DATABASE; do
+				echo "GRANT ALL ON $DATABASE.* TO '$MYSQL_USER'@'%' ;" >> "$TEMP_FILE"
+			done
 		fi
 	fi
 	
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -22,14 +22,18 @@ if [ -z "$(ls -A /var/lib/mysql)" -a "${1%_safe}" = 'mysqld' ]; then
 	EOSQL
 	
 	if [ "$MYSQL_DATABASE" ]; then
-		echo "CREATE DATABASE IF NOT EXISTS $MYSQL_DATABASE ;" >> "$TEMP_FILE"
+		for DATABASE in $MYSQL_DATABASE; do
+			echo "CREATE DATABASE IF NOT EXISTS $DATABASE ;" >> "$TEMP_FILE"
+		done
 	fi
 	
 	if [ "$MYSQL_USER" -a "$MYSQL_PASSWORD" ]; then
 		echo "CREATE USER '$MYSQL_USER'@'%' IDENTIFIED BY '$MYSQL_PASSWORD' ;" >> "$TEMP_FILE"
 		
 		if [ "$MYSQL_DATABASE" ]; then
-			echo "GRANT ALL ON $MYSQL_DATABASE.* TO '$MYSQL_USER'@'%' ;" >> "$TEMP_FILE"
+			for DATABASE in $MYSQL_DATABASE; do
+				echo "GRANT ALL ON $DATABASE.* TO '$MYSQL_USER'@'%' ;" >> "$TEMP_FILE"
+			done
 		fi
 	fi
 	


tfwright:
Hi all, I'm new to docker and trying out this image and I ran into a need to multiple databases. I've been sifting through this and other linked issues and for the life of me I can't parse what was settled on for best practice here. Do I need a new dockerfile that extends this one or can I specify multiple databases (or initdb scripts) in docker-compose.yml? Or something else entirely? Thanks for any guidance you can offer!

yosifkit:
@tfwright, you can have any number of initdb sql scripts, so that should be the easiest way to create multiple databases.

tfwright:
@yosifkit Thanks. To confirm, right now I'm just using the image in my compose config, so I'll need to change that to point to a custom dockerfile, right?

yosifkit:
@tfwright, not necessarily, you could just have the sql file with the database creation in the directory with the compose file and add a `volume` directive to the yaml with `./create-databases.sql:/docker-entrypoint-initdb.d/`. I think even `./*.sql:/docker-entrypoint-initdb.d/` should work if you have multiple scripts or even just put them in a folder: `./sql-setup/:/docker-entrypoint-initdb.d/`.

tfwright:
@yosifkit I tried adding a simple script and volume directive

```
#	new file:   src/sql-setup.sql
```
```
@@ -15,6 +15,7 @@ mysql:
   volumes:
      - "./src/my.cnf:/etc/mysql/conf.d/z_my.cnf"
      - /var/lib/mysql
+     - "./src/sql-setup.sql:/docker-entrypoint-initdb.d/"
```

 but docker complains when trying to start the container

```
Recreating 1632cce382_spectre_mysql_1...
Cannot start container 45b56877b97225ccd74c06bd3bfe09c7c06161b099922889ba0d4ba09bb0ad77: [8] System error: not a directory
```

What did I miss?

tfwright:
I was able to fix the above by changing the volume directive to "./src/sql-setup.sql:/docker-entrypoint-initdb.d/sql-setup.sql" which doesn't jive with what I understanding about docker's syntax...but who cares? It works!

Thanks again!

paulredmond:
@tfwright not sure if there is a better way, but your suggestion worked for me <3

solars:
Came across this ticket, but saw that it is still open. I'm using a mysql container for rails development like this:

docker run --name cpc-mysql -p 127.0.0.1:3307:3306 -e MYSQL_RANDOM_ROOT_PASSWORD=true -e MYSQL_USER=google_report -e MYSQL_PASSWORD=asdf -e MYSQL_DATABASE=app-dev -d mysql

I would need a second database app-test for testing, what is currently the cleanest way to do this?

thaJeztah:
@solars have you tried adding .sql or .sh file to the docker-entrypoint-initdb.d directory (or mounting a volume in that location)? https://github.com/docker-library/mysql/pull/90

solars:
Thanks for the pointer, I didn't see this option yet! In this case I will create a clean Dockerfile and add sql as needed. Thanks a lot!

solars:
@thaJeztah one more thing, I read here (first answer): http://stackoverflow.com/questions/32482780/how-to-create-populated-mysql-docker-image-on-build-time

that in the Dockerfile it doesn't work - is this still true? If so, how should I do it?

thaJeztah:
@solars correct; it's possible, but the [`VOLUME` declaration in the Dockerfile](https://github.com/docker-library/mysql/blob/ee6ac037ab647e0de9dbeb4e064610a95cb6df4a/5.7/Dockerfile#L40) complicates things. (also see https://github.com/docker/docker/pull/8177, https://github.com/docker/docker/issues/18286, https://github.com/docker/docker/issues/18287)

solars:
so I need to specify it in the run command above? or what is the suggested way to do it?

yosifkit:
Agreeing with https://github.com/docker-library/mysql/pull/86#issuecomment-188316391, I think this is solved.  Comment if you feel there is more to discuss.

Glideh:
So how do we do multiple databases after all ?
Separate container ? Seems to need quite a lot of config.
`MYSQL_DATABASE` space/coma separated ? Would be handy.

yosifkit:
@Glideh, either a new container or add a script to `/docker-entrypoint-initdb.d/`.  An `.sql` file would be the simplest, I think.

skytreader:
I'm trying to follow the discussion here but it seems that the script in `docker-entrypoint-initdb.d` is not being run at all.

My `Dockerfile` is as follows:

```
FROM ubuntu:16.04
ADD . ./app
WORKDIR ./app

RUN apt-get update
RUN apt-get install -y python python-pip
RUN pip install -r requirements.txt
```

Mounted in `docker-compose.yml` via

```
version: '2'
services:
  db:
    image: "mysql:5.7"
    environment:
      - MYSQL_ALLOW_EMPTY_PASSWORD=yes
    ports:
      - "3308:3306"
    volumes:
      - "./docker-setup/init.sql:/docker-entrypoint-initdb.d/init.sql"
  web:
    build: .
    command: python run.py
    ports:
      - "80:7070"
    volumes:
      - .:/app
    depends_on:
      - db
```

With `init.sql` containing

```
CREATE DATABASE IF NOT EXISTS app DEFAULT CHARACTER SET = utf-8;
CREATE DATABASE IF NOT EXISTS app_test DEFAULT CHARACTER SET = utf-8;
```

However, when I run `docker-compose up`, I ultimately end up with the error:

```
web_1  | OperationalError: (_mysql_exceptions.OperationalError) (1049, "Unknown database 'app'")
```

To be clear, using `MYSQL_DATABASE` env var allows my web and db to run harmoniously. It's just that, if possible, I want to run `app` and `app_test` on the same MySQL service.

Anything I'm doing wrong?

mrts:
> Anything I'm doing wrong?

Yes, you are attempting to mount files as volumes.

Here's a working example for PosgreSQL that has similar structure:

    myapp-postgresql:
        image: postgres:9.6.2
        volumes:
            - ../docker-posgresql-multi-db:/docker-entrypoint-initdb.d
        environment:
            - POSTGRES_MULTIPLE_DATABASES=db1,db2
            - POSTGRES_USER=myapp
            - POSTGRES_PASSWORD=

The `docker-posgresql-multi-db` directory contains a script for creating multiple databases, see https://github.com/mrts/docker-posgresql-multiple-databases for the script that you can reuse.

frankli0324:
I was encountering similar requirements lately, and I eventually found this.
to be honest, I'm not satisfied with the `docker-entrypoint-initdb.d` way around...
I'm thinking maybe there could be some environment variables with the "MYSQL_DATABASE" **PREFIX**, appended by "_0", "_1", "_2", etc.
for example, 
```yaml
- MYSQL_DATABASE=db1
- MYSQL_DATABASE_0=db2
- MYSQL_DATABASE_1=db3
# etc
```
it can be implemented within a loop in the entrypoint script
anyone want to have a try..? PRing the entry script

frankli0324:
I'll try sparing some time working with it

Issue #24, opened 2014-11-13
Provide mechanism to extend mysql image with additional SQL init scripts


yarosla:
Please provide mechanism to extend this image by providing additional SQL scripts to be executed upon initial server startup.

See official [postgres](https://registry.hub.docker.com/_/postgres/) image for example on how this can be done (“How to extend this image” section in readme).

You can arrange special directory for SQL files. All `*.sql` files found in that directory shall be appended to `$TEMP_FILE` that will be passed as `--init-file` parameter to MySQL.

johngian:
This would be really useful.
+1

jorgemoralespou:
It is not very usable without this feature. I have done as williamsdev in his PR for make it woking for me, but something from the official images should come, as it does in the postgres one.

ltangvald:
The changes we suggested in #53 should make this simpler to add, too, as the init sql is now just passed to the mysql client.

tophsic:
+1

Actually, I did a similar thing as @williamsbdev in #31

sherter:
+1
any news on this?

sweco-semhul:
+1

probepark:
+1

ltangvald:
This would be fairly simple to add with the new system for initializing the database used in #53.
However, that change is meant to be a temporary workaround, so it's probably best to do this in a way that doesn't depend on it.

tianon:
@ltangvald does that mean we're going to have a proper way to invoke instructions against `/var/lib/mysql` properly ala `postgres --single` without the silly limitations `--init-file` has? :smile:

roytruelove:
+1

PR #31, opened 2014-12-03
Feature: adding hook to execute additional scripts on startup

williamsbdev:
This will close both #18 and #24.

The scripts are located in the /docker-entrypoint-initdb.d directory.

Paired on this with @benkiefer

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -39,6 +39,8 @@ ENV PATH $PATH:/usr/local/mysql/bin:/usr/local/mysql/scripts
 WORKDIR /usr/local/mysql
 VOLUME /var/lib/mysql
 
+RUN mkdir -p /docker-entrypoint-initdb.d
+
 COPY docker-entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
 
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -33,6 +33,13 @@ if [ ! -d '/var/lib/mysql/mysql' -a "${1%_safe}" = 'mysqld' ]; then
 		fi
 	fi
 	
+	INIT_DIR='/docker-entrypoint-initdb.d'
+        for f in $INIT_DIR/*.sql
+        do 
+                echo "Appending sql script: $f"
+                cat $f >> "$TEMP_FILE"
+        done
+	
 	echo 'FLUSH PRIVILEGES ;' >> "$TEMP_FILE"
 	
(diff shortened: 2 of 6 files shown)


jgangemi:
is there any movement on this? i could really use the functionality this will provide.

yosifkit:
I think @tianon was working on a set closer to what we have in the postgres image (to include bash scripts as well as sql), but this will still be very limited and needs documentation to cover the caveats.

jgangemi:
is there another issue tracking this since you closed this one?

as a work around for anyone else, you could copy the entrypoint script that the container uses and append your init script contents to the end. if you're using `mysqldump` to create the file, be sure to use `--extended-insert=FALSE` and `--skip-comments`

williamsbdev:
There is not another issue. I was working through the comments but no
progress was made after I addressed the issues they had. I'm leaving it up
to someone else to do this if it is needed and will continue to just use my
fork.

On Saturday, January 24, 2015, Jae Gangemi <notifications@github.com> wrote:

> is there another issue tracking this since you closed this one?
>
> as a work around for anyone else, you could copy the entrypoint script
> that the container uses and append your init script contents to the end. if
> you're using mysqldump to create the file, be sure to use
> --extended-insert=FALSE and --skip-comments
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/docker-library/mysql/pull/31#issuecomment-71327453>.
>


-- 
Sent from Gmail Mobile

jgangemi:
then shouldn't it be enough to provide a link to the mysql docs for the `--init-file` option in the docs here and also include my comment above about the options to use for `mysqldump`?

if that's what it takes to get this into the official build, i'd be happy to create a pull request. if there's something else, please let me know and i can look to address it. having this as part of the official image would be extremely helpful to me (and i'm sure others).

williamsbdev:
Go ahead and try that. LGTM. I gave up as I felt they wanted me to research
how the Postgres was working and change my implementation. I have it
working for myself felt it was sufficient to be merged.

On Saturday, January 24, 2015, Jae Gangemi <notifications@github.com> wrote:

> then shouldn't it be enough to provide a link to the mysql docs for the
> --init-file option in the docs here and also include my comment above
> about the options to use for mysqldump?
>
> if that's what it takes to get this into the official build, i'd be happy
> to create a pull request. if there's something else, please let me know and
> i can look to address it. having this as part of the official image would
> be extremely helpful to me (and i'm sure others).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/docker-library/mysql/pull/31#issuecomment-71328843>.
>


-- 
Sent from Gmail Mobile

PR #33, opened 2014-12-10
improve the start speed of the container 

benkiefer:
We were noticing a significant decrease in container start time compared to when we were using the postgres container (20 seconds w/mysql vs 2 seconds) and started to do some debugging to figure out where the pain was. It turns out it came mostly from 'mysql_install_db' step in the entrypoint. Since this container is opinionated about always using the mysql user anyway, it seemed to make sense to move this to the initial build of the container, rather than as part of the run.

The container now starts up in 2-3 seconds. 

I'd like to know what you think of this, and if there were any reasons that the mysql_install_db step was part of the entrypoint vs part of the container build.

I only applied this change to 5.7 because I wanted to talk about it before going down the route of making this change to all three versions.

Thanks for the feedback.

### DIFF
--- a/5.7/Dockerfile
+++ b/5.7/Dockerfile
@@ -37,9 +37,14 @@ RUN apt-get update && apt-get install -y curl --no-install-recommends && rm -rf 
 ENV PATH $PATH:/usr/local/mysql/bin:/usr/local/mysql/scripts
 
 WORKDIR /usr/local/mysql
+RUN mysql_install_db --user=mysql --datadir=/var/lib/mysql
+
+RUN chown -R mysql:mysql /var/lib/mysql
+
 VOLUME /var/lib/mysql
 
 COPY docker-entrypoint.sh /entrypoint.sh
+
 ENTRYPOINT ["/entrypoint.sh"]
 
 EXPOSE 3306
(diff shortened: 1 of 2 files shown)


rsandila:
What if you:
1.  Created some "marker" file in /var/lib/mysql in Dockerfile.
2. Check if that file exists in docker-entrypoint.sh and if it does skip mysql_install_db otherwise run it?

yosifkit:
So I did a test where I did the `initialize-insecure` within the image and then copied it to the volume on start if the volume was empty.  I did not see a significant improvement over the regular startup time, but the image was now ~26MB bigger :cry:.

@ltangvald, Are there some settings in the configuration file that would change what mysql creates with `mysql_install_db`/`--initialize-insecure`?

ltangvald:
I think this would give a bigger benefit for 5.5 and 5.6. The 5.7 server takes quite a lot longer to start than previous versions, so skipping --initialize wouldn't give as big a boost.

ltangvald:
Do you know how common the use case is? Is the reduction in start up time the first time the image is used worth the added size?

tianon:
I think the conclusion here is that the benefit isn't worth the added image size (especially given the necessary complexity increase).  Thanks for opening the discussion though!

PR #39, opened 2015-01-08
cleanup: add escaping of MYSQL_DATABASE to support hyphenated database names

PeterKneale:
In it's current form, hyphenated database names are not supported.
e.g. with a database name of "Products-Staging"

The following error is reported when creating the database:
```
ERROR: 1064  You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-Staging ;' at line 1
```
The following error is reported when granting permission to the specified user:
```
ERROR: 1046  No database selected
```
Adding the escaping around the MYSQL_DATABASE variable fixes this.


tianon:
LGTM

yosifkit:
LGTM

PR #42, opened 2015-01-16
update entrypoint to read datadir from config file

semarj:


yosifkit:
I am hesitant since we explicitly define this directory as a volume in the [Dockerfile](https://github.com/docker-library/mysql/blob/master/5.7/Dockerfile#L35).  Is there a compelling reason to need to specify a different data directory?

semarj:
We are using the mysql base image to build a test mysql image with data pre-packaged (via jenkins jobs).  Since /var/lib/mysql is a volume in the Dockerfile, the data saved to the database is not persisted in the built image.  This was our experience at least.  Is that not what you would have expected?

yosifkit:
That seems valid to me. What do you think @tianon?

thaJeztah:
What worries me is that @semarj is on *purpose* not using a volume for the data-dir, but other people may set `DATADIR`, not knowing that the data-dir will end up in the container (not in a volume).

If you really need to override the directory, you'd still be able to, using something like;

    FROM mysql
    RUN sed -i -e 's#/var/lib/mysql#/my/custom/path#' /entrypoint.sh

tianon:
I'm actually -1 on this particular patch, especially since it's only half the story as-is:
```console
$ docker run --rm mysql grep -rn /var/lib/mysql /etc/mysql/
/etc/mysql/my.cnf:37:datadir		= /var/lib/mysql
```

This is why I'd much rather see us scrape this information directly from the MySQL config (as my original `TODO` comment stated), since that's a much more upstream-friendly fix.

PR #43, opened 2015-01-22
Support for basic Master-Slave Replication on 5.6

ioggstream:
Actual images does not support replication parameters.

This patch supports the following variables:
+#  - MYSQL_ROOT_PASSWORD
+#  - MYSQL_REPLICA_USER: create the given user on the intended master host
+#  - MYSQL_REPLICA_PASS
+#  - MYSQL_MASTER_SERVER: change master on this location on the intended slave
+#  - MYSQL_MASTER_PORT: optional, by default 3306
+#  - MYSQL_MASTER_WAIT_TIME: seconds to wait for the master to come up

You can test the resulting image with the fig.yml present here https://registry.hub.docker.com/u/ioggstream/mysql/

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -71,7 +71,7 @@ if [ ! -d "$DATADIR/mysql" -a "${1%_safe}" = 'mysqld' ]; then
 	#
         if [ "$MYSQL_MASTER_SERVER" ]; then
                 MYSQL_MASTER_PORT=${MYSQL_MASTER_PORT:-3306}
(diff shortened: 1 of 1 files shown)


yosifkit:
I'll defer to @ltangvald as to whether we want to add such functionality to the image.

Closing old issue, if you feel this was closed in error please let us know.

ltangvald:
I think it's better to handle this with the init system.

Though as these fairly common use-cases keep coming in maybe we could have an examples section in the repo with some simple scripts that demonstrate how to do them.

PR #46, opened 2015-01-29
Update entrypoints to read DATADIR from the MySQL configuration directly instead of assuming /var/lib/docker

### LINKED ISSUE #40: How to Populate database at container creation
I have the container running, all the MYSQL_ parameters worked, the database was created and the user was defined and granted access to the database.. running in daemon mode

the database is empty as expected
I want to execute the sql file that contains all the DB schema commands (create table, ....)

I can create a script and run docker exec mysql, 
IF I did -v to map some host volume into the container when I started it

is this the proper way? 

in my script, how do I wait til the daemon container is up and running?
try docker exec repeatedly til it works? (shouldn't work til the container CMD is running)

I have also created my own image which does all the same things this does and adds support for database population, but it makes no sense to manage my own image.

tianon:
Don't be afraid -- most of the diff is reindentation!

This will actually let `datadir` be defined anywhere in the config (even in `conf.d` files and the command line via `docker run <image> mysqld --datadir=/somewhere/else`, which is how I was primarily testing the change) because we're letting MySQL itself tell us what the value should be.

Fixes #40
Closes #42

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -1,52 +1,55 @@
 #!/bin/bash
 set -e
 
-# TODO read this from the MySQL config?
-DATADIR='/var/lib/mysql'
-
 if [ "${1:0:1}" = '-' ]; then
 	set -- mysqld "$@"
 fi
 
-if [ ! -d "$DATADIR/mysql" -a "${1%_safe}" = 'mysqld' ]; then
-	if [ -z "$MYSQL_ROOT_PASSWORD" -a -z "$MYSQL_ALLOW_EMPTY_PASSWORD" ]; then
-		echo >&2 'error: database is uninitialized and MYSQL_ROOT_PASSWORD not set'
-		echo >&2 '  Did you forget to add -e MYSQL_ROOT_PASSWORD=... ?'
-		exit 1
-	fi
+if [ "$1" = 'mysqld' ]; then
+	# read DATADIR from the MySQL config
+	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
 	
-	echo 'Running mysql_install_db ...'
-	mysql_install_db --basedir=/usr/local/mysql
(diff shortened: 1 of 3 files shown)


yosifkit:
LGTM

md5:
I wish Github would just support the damn `?w=1` feature in the UI: https://github.com/docker-library/mysql/pull/46/files?w=1

md5:
Also, LGTM :metal:

tianon:
Oooh, that's sexy @md5 :D

thaJeztah:
> I wish Github would just support the damn ?w=1 feature in the UI: https://github.com/docker-library/mysql/pull/46/files?w=1

Wot?! They have that?! Awesome!

md5:
It's mentioned here along with some other not-as-useful "secrets": https://github.com/blog/967-github-secrets

tianon:
Yay, I already knew the rest (ie, I'm not missing out on any more secret
cool features that I neeeeed, because I neeeeeds it)! \o/

psychemedia:
Does `--datadir=/somewhere/else` run equally with `mysqld_safe`?

tianon:
What's the use-case for using "mysqld_safe" with this image?

The "--verbose --help" trick we use here to scrape the "datadir" value
directly from mysqld itself doesn't work with "mysqld_safe", but the
benefits of using "mysqld_safe" here seem dubious to me, so I'm curious
what that gains.

psychemedia:
@tianon Way past the limits of my cut and paste knowledge. I keep seeing a pattern:

   ADD init.sql /tmp/
   RUN (/usr/bin/mysqld_safe &); sleep 3; mysql -u root < /tmp/init.sql

for trying to get databases initialised. The use case I am trying to find a solution to is described in https://github.com/docker-library/mysql/issues/40#issuecomment-72107997 with additional straw clutching at http://stackoverflow.com/questions/28227755/can-i-use-fig-to-initialise-a-persisting-database-in-docker

PR #50, opened 2015-02-11
Fix issue in Entrypoint

fchaillou:
- The datadir is retrieve from the config, we don't need to add /mysql to the result
- This induces a myslq_install_db at each docker run which is not the intended behavior

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -9,7 +9,7 @@ if [ "$1" = 'mysqld' ]; then
 	# read DATADIR from the MySQL config
 	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
 	
-	if [ ! -d "$DATADIR/mysql" ]; then
+	if [ ! -d "$DATADIR" ]; then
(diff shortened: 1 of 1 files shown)


tianon:
I'm not sure I see the problem -- this is on purpose; any properly-initialized MySQL data directory _will_ have a mysql directory inside it which corresponds to the `mysql` database: (which is why we check for this directory existing instead of just checking if the whole directory is empty like we used to)
```console
$ docker run -it --rm mysql bash
root@a164f399271f:/# ls -l /var/lib/mysql 
total 0
root@a164f399271f:/# mysql_install_db --datadir=/var/lib/mysql &> /dev/null
root@a164f399271f:/# ls -l /var/lib/mysql
total 110592
-rw-rw---- 1 mysql mysql 50331648 Feb 11 16:38 ib_logfile0
-rw-rw---- 1 mysql mysql 50331648 Feb 11 16:38 ib_logfile1
-rw-rw---- 1 mysql mysql 12582912 Feb 11 16:38 ibdata1
drwx------ 1 mysql mysql     2486 Feb 11 16:38 mysql
drwx------ 1 mysql mysql     3296 Feb 11 16:38 performance_schema
```

fchaillou:
Yeah you're right,

We have an issue with our base being recreated at each run and it seemed like it was the issue.
I will investigate more then  !
Thanks for your quick reply !
fabien

PR #53, opened 2015-03-06
Prevent root password from being accessible by normal users on first run

ltangvald:
The first time the server was started, the file containing the root password in plaintext would be readable by any user on the database (and the filename listed in the variable list).

We changed the first run to do the initialization separately, then restart the server.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -1,14 +1,25 @@
 #!/bin/bash
 set -e
 
+get_option () {
+    local section=$1
+    local option=$2
+    local default=$3
+    ret=$(/usr/bin/my_print_defaults $section | grep '^--'${option}'=' | cut -d= -f2-)
+    [ -z $ret ] && ret=$default
+    echo $ret
+}
+
 # if command starts with an option, prepend mysqld
 if [ "${1:0:1}" = '-' ]; then
 	set -- mysqld "$@"
 fi
 
 if [ "$1" = 'mysqld' ]; then
-	# read DATADIR from the MySQL config
-	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
+	# Get config
+        DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
+        SOCKET=$(get_option  mysqld socket "$datadir/mysql.sock")
+        PIDFILE=$(get_option mysqld pid-file "/var/run/mysqld/mysqld.pid")
 	
 	if [ ! -d "$DATADIR/mysql" ]; then
 		if [ -z "$MYSQL_ROOT_PASSWORD" -a -z "$MYSQL_ALLOW_EMPTY_PASSWORD" ]; then
@@ -16,16 +27,27 @@ if [ "$1" = 'mysqld' ]; then
 			echo >&2 '  Did you forget to add -e MYSQL_ROOT_PASSWORD=... ?'
 			exit 1
 		fi
-		
-		echo 'Running mysql_install_db ...'
-		mysql_install_db --datadir="$DATADIR" --basedir=/usr/local/mysql
+		mkdir -p $DATADIR
+		echo 'Running mysql_install_db'
+		mysql_install_db --user=mysql --datadir=$DATADIR --rpm
 		echo 'Finished mysql_install_db'
-		
+
+		mysqld --user=mysql --datadir=$DATADIR --skip-networking &
+                for i in $(seq 30 -1 0); do
+		    [ -S $SOCKET ] && break
+		    echo 'MySQL init process in progress...'
+		    sleep 1
+		done
+                if [ $i = 0 ]; then
+                    echo >&2 'MySQL init process failed.'
+                    exit 1
+                fi
+                
 		# These statements _must_ be on individual lines, and _must_ end with
 		# semicolons (no line breaks or comments are permitted).
 		# TODO proper SQL escaping on ALL the things D:
 		
-		tempSqlFile='/tmp/mysql-first-time.sql'
+		tempSqlFile=$(mktemp /tmp/mysql-first-time.XXXXXX.sql)
 		cat > "$tempSqlFile" <<-EOSQL
 			-- What's done in this file shouldn't be replicated
 			--  or products like mysql-fabric won't work
@@ -38,23 +60,35 @@ if [ "$1" = 'mysqld' ]; then
 		EOSQL
 		
 		if [ "$MYSQL_DATABASE" ]; then
-			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ;" >> "$tempSqlFile"
+			echo "CREATE DATABASE IF NOT EXISTS \`"$MYSQL_DATABASE"\` ;" >> "$tempSqlFile"
 		fi
 		
 		if [ "$MYSQL_USER" -a "$MYSQL_PASSWORD" ]; then
-			echo "CREATE USER '$MYSQL_USER'@'%' IDENTIFIED BY '$MYSQL_PASSWORD' ;" >> "$tempSqlFile"
+			echo "CREATE USER '"$MYSQL_USER"'@'%' IDENTIFIED BY '"$MYSQL_PASSWORD"' ;" >> "$tempSqlFile"
 			
 			if [ "$MYSQL_DATABASE" ]; then
-				echo "GRANT ALL ON \`$MYSQL_DATABASE\`.* TO '$MYSQL_USER'@'%' ;" >> "$tempSqlFile"
+				echo "GRANT ALL ON \`"$MYSQL_DATABASE"\`.* TO '"$MYSQL_USER"'@'%' ;" >> "$tempSqlFile"
 			fi
 		fi
 		
 		echo 'FLUSH PRIVILEGES ;' >> "$tempSqlFile"
 		
-		set -- "$@" --init-file="$tempSqlFile"
+		chown -R mysql:mysql "$DATADIR"
+		mysql -uroot < $tempSqlFile
+
+		rm -f $tempSqlFile
+		kill $(cat $PIDFILE)
+                for i in $(seq 30 -1 0); do
+		    [ -S $SOCKET ] || break
+		    echo 'MySQL init process in progress...'
+		    sleep 1
+		done
+                if [ $i = 0 ]; then
+                    echo >&2 'MySQL hangs during init process.'
+                    exit 1
+                fi
+		echo 'MySQL init process done. Ready for start up.'
 	fi
-	
-	chown -R mysql:mysql "$DATADIR"
 fi
 
 exec "$@"
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -20,7 +20,7 @@ if [ "$1" = 'mysqld' ]; then
 	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
 	SOCKET=$(get_option  mysqld socket "/tmp/mysql.sock")
 	HOSTNAME=$(hostname)
-	PIDFILE=$(get_option mysqld pid-file "$DATADIR/$HOSTNAME.pid")
+	PIDFILE=$(get_option mysqld pid-file "$DATADIR/mysqld.pid")
 
 	if [ ! -d "$DATADIR/mysql" ]; then
 		if [ -z "$MYSQL_ROOT_PASSWORD" -a -z "$MYSQL_ALLOW_EMPTY_PASSWORD" ]; then
@@ -28,6 +28,7 @@ if [ "$1" = 'mysqld' ]; then
 			echo >&2 '  Did you forget to add -e MYSQL_ROOT_PASSWORD=... ?'
 			exit 1
 		fi
+
 		mkdir -p "$DATADIR"
 		chown -R mysql:mysql "$DATADIR"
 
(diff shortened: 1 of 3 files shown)


5.5/docker-entrypoint.sh
```diff
@@ -33,23 +58,35 @@ if [ "$1" = 'mysqld' ]; then
...
+		    sleep 1
+		done
+                if [ $i = 0 ]; then
+                    echo >&2 'MySQL hangs during init process.'
+                    exit 1
+                fi
+		echo 'MySQL init process done. Ready for start up.'
 	fi
-	
-	chown -R mysql:mysql "$DATADIR"
```
thaJeztah: Permissions are now only set during initialization. Would this be a problem when using an already initialized data-dir (e.g. bind-mounted)?

ltangvald: Yes, it almost certainly will. The line should have been copied instead of moved. Thanks :)

md5:
If I understand this correctly, the issue is that the file `/tmp/mysql-first-time.sql` can be read by any user who can connect to the database by using `LOAD DATA INFILE`. In order to do that, the user would have to have the `FILE` permission granted on some table in the database. Is that right?

I tried changing the permissions on the file to `0400` in the hope that MySQL would read it as `root` before dropping privileges to the `mysql` user, but that was not the case.

One thing I don't like about the entrypoint changes is the fact that `mysqld` is started and backgrounded in the middle of the initialization. It would be good if there were a more graceful way to achieve the same fix.

On a side note, it looks like the branch you're using has other changes in it that are unrelated to the password fix (e.g. the removal of Perl for the 5.7 Dockerfile). Those changes should probably go in a separate PR.

md5:
P.S. It's really cool to see this PR coming from the `mysql/docker` repo. It would be great to see more official images taken over by the upstream teams.

ltangvald:
Is the desired procedure to only have a single commit, or just no unrelated features? The perl dependency change could be removed, but the others are bug fixes.

With regards to the server run in init, I agree it's fairly ugly, but don't really know a more graceful way to do it. We do have suggestions on features that would help, such as single user mode, but the server doesn't currently support it.

md5:
I mainly suggested breaking out the Perl change because it looks like something that can be merged without much discussion, whereas I think sorting out this `MYSQL_ROOT_PASSWORD` thing may take a bit more time (given stuff like the backgrounded `mysqld`).

If you think any of the other fixes are unambiguous bugs that can be easily merged after a short discussion, I'd probably break those into their own PRs too (or possibly combine them with the Perl change).

ltangvald:
The bugfixes are related to the security issue, so they need to be included. We'll separate the Perl change, though.

ltangvald:
Removed the perl change.

While the system of starting and restarting the server is clunky, it does have the benefit of making it easier to support features such as those requested in issue #24 and pull request #18 without using --bootstrap or --init-file, which have their own problems.

5.5/docker-entrypoint.sh
```diff
@@ -1,13 +1,25 @@
 #!/bin/bash
 set -e
 
+get_option () {
+    local section=$1
+    local option=$2
+    local default=$3
+    ret=$(/usr/local/mysql/bin/my_print_defaults $section | grep '^--'${option}'=' | cut -d= -f2-)
```
tianon: Why not just rely on the `PATH` for finding the right `my_print_defaults` binary?

5.5/docker-entrypoint.sh
```diff
@@ -1,13 +1,25 @@
...
+
 if [ "${1:0:1}" = '-' ]; then
 	set -- mysqld "$@"
 fi
 
 if [ "$1" = 'mysqld' ]; then
-	# read DATADIR from the MySQL config
-	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
+	# Get config
+        DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
```
tianon: Shouldn't this be using the new `get_option` function too?

5.5/docker-entrypoint.sh
```diff
@@ -1,13 +1,25 @@
...
+        SOCKET=$(get_option  mysqld socket "/tmp/mysql.sock")
+        HOSTNAME=$(hostname)
+        PIDFILE=$(get_option mysqld pid-file "$DATADIR/$HOSTNAME.pid")
```
tianon: Why is this default value hostname-dependent?

5.5/docker-entrypoint.sh
```diff
@@ -33,22 +58,36 @@ if [ "$1" = 'mysqld' ]; then
...
 			if [ "$MYSQL_DATABASE" ]; then
-				echo "GRANT ALL ON \`$MYSQL_DATABASE\`.* TO '$MYSQL_USER'@'%' ;" >> "$tempSqlFile"
+				echo "GRANT ALL ON \`"$MYSQL_DATABASE"\`.* TO '"$MYSQL_USER"'@'%' ;" >> "$tempSqlFile"
 			fi
 		fi
 		
 		echo 'FLUSH PRIVILEGES ;' >> "$tempSqlFile"
 		
-		set -- "$@" --init-file="$tempSqlFile"
+		chown -R mysql:mysql "$DATADIR"
```
tianon: Isn't it kind of dangerous to be doing this `chown` while `mysqld` is actively running and using `$DATADIR`?  Shouldn't this be up before we start `mysqld` instead?

tianon:
+1 to being able to solve things like #24 in a clean way -- this was the solution we were considering for that, but were having trouble bringing ourselves to actually start up and background `mysqld` just to talk to it for a little bit, shut it down, and then start it back up again.

tianon:
The whitespace in this PR seems to be all over the place -- some places it's a tab (like it was originally in these files), but some places it's a 4-space indent or stranger incantations. :smile:

ltangvald:
Whitespace issue is my bad. Managed to set up my editor and git so I missed the issue :)

5.5/docker-entrypoint.sh
ltangvald: We noticed some examples of people passing --datadir to the container when starting it (i.e. to save data in a location that would be included when committing the container). get_option only fetches the default values, so this would no longer work.

5.5/docker-entrypoint.sh
ltangvald: Yes, that should work better.

5.5/docker-entrypoint.sh
ltangvald: The pidfile option isn't included in the mysql config of the package used by this image.
We could append the option to the config file, but this seemed as clean a solution.

5.5/docker-entrypoint.sh
ltangvald: Yes, the ownership should be ensured before the init process is started.

5.5/docker-entrypoint.sh
```diff
@@ -1,54 +1,94 @@
...
 fi
 
 if [ "$1" = 'mysqld' ]; then
-	# read DATADIR from the MySQL config
+	# Get config
 	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
-	
+	SOCKET=$(get_option  mysqld socket "/tmp/mysql.sock")
+	HOSTNAME=$(hostname)
+	PIDFILE=$(get_option mysqld pid-file "$DATADIR/$HOSTNAME.pid")
```
tianon: What I meant with my comment on this line before is why are we bothering with `HOSTNAME` here at all?  The "hostname" in our container is likely to be something random like `9d3cb2141e44`, so why not just go with something like `mysql.pid` and make sure it's deleted appropriately before we start up `mysqld`?

5.5/docker-entrypoint.sh
```diff
@@ -1,54 +1,94 @@
...
 	if [ ! -d "$DATADIR/mysql" ]; then
 		if [ -z "$MYSQL_ROOT_PASSWORD" -a -z "$MYSQL_ALLOW_EMPTY_PASSWORD" ]; then
 			echo >&2 'error: database is uninitialized and MYSQL_ROOT_PASSWORD not set'
 			echo >&2 '  Did you forget to add -e MYSQL_ROOT_PASSWORD=... ?'
 			exit 1
 		fi
-		
-		echo 'Running mysql_install_db ...'
-		mysql_install_db --datadir="$DATADIR" --basedir=/usr/local/mysql
+		if [ ! -d "$DATADIR" ]; then
```
tianon: `mkdir -p` will gracefully handle the case of `$DATADIR` already existing; that's the whole point of `-p` :smile:

5.5/docker-entrypoint.sh
```diff
@@ -1,54 +1,94 @@
...
+			mkdir -p $DATADIR
```
tianon: `$DATADIR` here ought to be quoted, as it is below in the `chown`

5.5/docker-entrypoint.sh
```diff
@@ -1,54 +1,94 @@
...
+		fi
+		chown -R mysql:mysql "$DATADIR"
+
+		echo 'Running mysql_install_db'
+		mysql_install_db --user=mysql --datadir=$DATADIR --rpm --basedir=/usr/local/mysql
```
tianon: `$DATADIR` here ought to be quoted, as it is above in the `chown`

5.5/docker-entrypoint.sh
```diff
@@ -1,54 +1,94 @@
...
 			DELETE FROM mysql.user ;
 			CREATE USER 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
 			GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION ;
 			DROP DATABASE IF EXISTS test ;
 		EOSQL
-		
+
 		if [ "$MYSQL_DATABASE" ]; then
-			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ;" >> "$tempSqlFile"
+			echo "CREATE DATABASE IF NOT EXISTS \`"$MYSQL_DATABASE"\` ;" >> "$tempSqlFile"
```
tianon: Since this is already inside double quotes, why bother with ending them to drop in the variable? (alternatively, why not switch to single quotes instead so that the backtick doesn't have to be escaped?)

5.7/docker-entrypoint.sh
```diff
@@ -1,55 +1,97 @@
...
-		
-		echo 'Running mysql_install_db ...'
-		mysql_install_db --datadir="$DATADIR" --mysqld-file="$(which "$1")"
+		if [ ! -d "$DATADIR" ]; then
+			mkdir -p $DATADIR
+		fi
+		chown -R mysql:mysql "$DATADIR"
+
+		echo 'Running mysql_install_db'
+		mysql_install_db --user=mysql --datadir=$DATADIR --insecure --mysqld-file=/usr/sbin/mysqld
```
tianon: This line was much improved in https://github.com/docker-library/mysql/pull/56/files#diff-17ed820d5141a5e8565a0431e5385daeR20 :+1: :smile:

5.7/docker-entrypoint.sh
```diff
@@ -1,55 +1,97 @@
...
+		done
+		if [ $i = 0 ]; then
+			echo >&2 'MySQL init process failed.'
+			exit 1
+		fi
+
+		# Workaround for bug in 5.7 that doesn't clean up after itself correctly 
+		rm $DATADIR/ib_logfile0
+		rm $DATADIR/ib_logfile1
+		rm $DATADIR/ibdata1
```
tianon: Is this still relevant with 5.7.6-m16?  Also, shouldn't we wait to delete these until `mysqld` isn't running anymore? (because this is happening right after we've started `mysqld` right now)

tianon: And the follow up question, is there a bug filed for these upstream that we can link to for an easier way to check when it's fixed? :smile:

ltangvald: This needs to be changed for 5.7.6. There's a similar behavior when running mysqld --verbose --help which will cause mysqld --initialize to fail because of these files being created. http://bugs.mysql.com/bug.php?id=75995 is the bug (though from a slightly different perspective).
However, using --verbose --help to get datadir as we do is something of a hack, and we want to rewrite this part a bit so we'll no longer need a hack to fix issues with another hack :)

5.5/docker-entrypoint.sh
ltangvald: Good point. I didn't remember it does more than create parent directories :)

5.5/docker-entrypoint.sh
ltangvald: Will go through and fix the inconsistent quoting. Thanks

5.7/docker-entrypoint.sh
ltangvald: Will replace it with the one from the other PR :)

5.5/docker-entrypoint.sh
ltangvald: Yeah, it's probably a better idea to specify the pid-file instead of doing it like this.

5.7/docker-entrypoint.sh
ltangvald: The workaround might not be needed for the debian-based image, though, since the apt package doesn't set ownership on /var/lib/mysql to the mysql user.

tianon: :heart_eyes: sounds awesome :+1:

ltangvald:
The 5.7.6+ issue wasn't given very high priority before we started the work with Docker, but it's a bigger problem here because of the way we use the --verbose --help command to fetch options. We're looking into ways to handle it, but at least it's not an issue on Debian right now :)

tianon:
Good enough for me -- let's get this in; sorry for the delay!

LGTM

yosifkit:
Can you rebase so that the pretty new tests can run (#73)?  Just a `git pull --rebase https://github.com/docker-library/mysql.git master` and then `git push -f`.

ltangvald:
Will do

ltangvald:
A couple of things to note:
- Adding --innodb-read-only to the --verbose --help command prevents it from writing to an empty datadir, should this become an issue on Debian as well.
- We're working on adding functionality that will remove the need for this whole -start-stop-check-if-stopped weirdness, so should be able to improve this in the future :)

tianon:
LGTM

(YOU TEASE -- looking forward to something better than our start-stop hackery :+1:)

tianon:
Also, green tests are :green_heart:.

yosifkit:
LGTM: https://travis-ci.org/docker-library/mysql/jobs/65208290#L611 :smiley:

PR #60, opened 2015-04-02
fix: don't replicate local setup

ioggstream:
The following patch:
 - avoid replicating local user setup to slave hosts;
 - pass --defaults-file to mysql_install_db

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 
+# if command starts with an option, prepend mysqld
 if [ "${1:0:1}" = '-' ]; then
 	set -- mysqld "$@"
 fi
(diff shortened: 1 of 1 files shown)


5.6/docker-entrypoint.sh
```diff
@@ -26,6 +29,10 @@ if [ "$1" = 'mysqld' ]; then
 		
 		tempSqlFile='/tmp/mysql-first-time.sql'
 		cat > "$tempSqlFile" <<-EOSQL
+		        -- What's done in this file shouldn't be replicated
```
tianon: Mixed tabs and spaces here. :wink:

5.6/docker-entrypoint.sh
```diff
@@ -17,7 +18,9 @@ if [ "$1" = 'mysqld' ]; then
 		fi
 		
 		echo 'Running mysql_install_db ...'
-		mysql_install_db --datadir="$DATADIR"
+		# pass --defaults-file and --defaults-extra-file to mysql_install_db
+		DEFAULTS_FILE_ARGS=$(egrep -o -- '--defaults(|-extra)-file=[^ ]+' <<< "$@")
+		mysql_install_db --datadir="$DATADIR" $DEFAULTS_FILE_ARGS
```
tianon: This change is a lot more fragile (and thus controversial) than the `SET @@SESSION.SQL_LOG_BIN` is.

ioggstream: Thx @tianon for your review!

You don't like the code or the functionality?

5.6/docker-entrypoint.sh
ioggstream: fixing...(I got kate in python mode)

yosifkit:
LGTM

tianon:
LGTM

PR #61, opened 2015-04-03
feature: pass defaults-file to mysql_install_db to honor innodb_log_file...

ioggstream:
Now mysql_install_db does not honor innodb_log_file_size provided in my.cnf.

This patch should fix it up.

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 
+# if command starts with an option, prepend mysqld
 if [ "${1:0:1}" = '-' ]; then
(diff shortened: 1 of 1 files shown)


yosifkit:
Sorry for the slow reply, is this still a needed feature?  ie, able to pass arguments to the `mysql_install_db`/`mysqld --initialize-insecure=on`.

cc @ltangvald

ltangvald:
Couldn't this be handled just by mapping up your extra config file to one of the regular config locations?

tianon:
Closing, given @ltangvald and @yosifkit's notes above.  This should likely be handled via dropping a file in the correct location or via `mysqld` arguments supplied as part of the container command.

PR #75, opened 2015-06-11
Make get_option() handle duplicate options in my_print_defaults output

### LINKED ISSUE #74: "/entrypoint.sh: line 9: [: /var/run/mysqld/mysqld.sock: binary operator expected"
docker-compose.yml:
```yaml
web:
  build: .
  ports:
    - "80:80"
  links:
    - db
db:
  build: docker/db

```

docker/db/Dockerfile:
```Dockerfile
FROM mysql:5.6
MAINTAINER Mozilla

ENV MYSQL_ROOT_PASSWORD=root \
    MYSQL_DATABASE=bugs \
    MYSQL_USER=bugzilla \
    MYSQL_PASSWORD=bugs

EXPOSE 3606

VOLUME /var/lib/mysql

COPY my.cnf /etc/mysql/conf.d/
RUN chmod 644 /etc/mysql/conf.d/my.cnf
```

docker-compose up:
```bash
Recreating bmo_db_1...
Creating bmo_db_1...
Recreating bmo_web_1...
Creating bmo_web_1...
Attaching to bmo_db_1, bmo_web_1
db_1  | /entrypoint.sh: line 9: [: /var/run/mysqld/mysqld.sock: binary operator expected
db_1  | 2015-06-07 00:01:47 0 [Note] mysqld (mysqld 5.6.25) starting as process 1 ...
```

Am I doing something wrong, or is the "binary operator expected" warning/error a bug?

```bash
docker@boot2docker:/c/Users/Ed/src/bmo$ docker --version
Docker version 1.6.2, build 7c8fca2
docker@boot2docker:/c/Users/Ed/src/bmo$ docker-compose --version
docker-compose 1.3.0rc1
```

Thanks! :-)

edmorley:
get_option() was added by #53 to extract `socket` and `pid-file` from my_print_defaults. However if a custom mysql configuration file has been added, my_print_defaults returns duplicates (it merely lists all options specified in any config file and doesn't de-dupe). In these cases, `ret` would equal:
`/var/run/mysqld/mysqld.sock /var/lib/mysql/mysqld.sock`
...causing:
`/entrypoint.sh: line 9: [: /var/run/mysqld/mysqld.sock: binary operator expected`

As such, get_option() should just return the last matching option found in the my_print_defaults output - which should correspond to the one in the custom configuration file (rather than say the global default).

Fixes #74.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -5,7 +5,9 @@ get_option () {
 	local section=$1
 	local option=$2
 	local default=$3
-	ret=$(my_print_defaults $section | grep '^--'${option}'=' | cut -d= -f2-)
+	# my_print_defaults can output duplicates, if an option exists both globally and in
+	# a custom config file. We pick the last occurence, which is from the custom config.
+	ret=$(my_print_defaults $section | grep '^--'${option}'=' | cut -d= -f2- | tail -n1)
 	[ -z $ret ] && ret=$default
 	echo $ret
 }
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -5,7 +5,9 @@ get_option () {
 	local section=$1
 	local option=$2
 	local default=$3
-	ret=$(my_print_defaults $section | grep '^--'${option}'=' | cut -d= -f2-)
+	# my_print_defaults can output duplicates, if an option exists both globally and in
+	# a custom config file. We pick the last occurence, which is from the custom config.
+	ret=$(my_print_defaults $section | grep '^--'${option}'=' | cut -d= -f2- | tail -n1)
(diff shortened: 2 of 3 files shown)


edmorley:
Travis is failing, however when I reverted my change for mysql:5.6, and pushed again, I got the same failure. In addition, the mysql:5.7 failure is a timeout during the build, which I can't see how is related to an entrypoint script change. As such, I believe the Travis runs are broken on master due to changes in https://github.com/docker-library/official-images or elsewhere.

I've filed #76 and #77 to track them separately.

tianon:
LGTM

(and I'd personally be +1 for classifying this as a bug :+1:)

edmorley:
Thanks! Will file a bug again mysql's my_print_defaults when I get a chance :-)

Issue #69, opened 2015-05-03
MySQL docker 5.7.6 and later fails to initialize database


guss77:
When using the docker image to start without an existing database, the container's entrypoint.sh script tries to call the `mysqld` binary to create the database. This fails in versions later than 5.7.5 because the script starts by calling `mysqld --verbose --help` to get the configured datadir and when that runs and there is no database, it initializes it automatically. 

When the script then calls `--initialize`, that fails with the error:

     [ERROR] --initialize specified but the data directory has files in it. Aborting.

The workaround is to run mysql:5.7.5 to create the database, then upgrade to a later image, which will work because the database already exists.

md5:
The `mysql:5.7.6` image was tested with this test before release: https://github.com/docker-library/official-images/blob/master/test/tests/mysql-basics/run.sh

You may need to provide some more information about your failure.

guss77:
The problem is one of permissions: the container has the data dir `/var/lib/mysql` owned by root, so when `entrypoint.sh` is calling `mysqld --help` to get the data dir, the directory isn't writable by mysqld and mysqld can't write the innodb log files it will normally create automatically when there is no database already set up. `entrypoint.sh` will later chown the directory.

The problem is that the container relies on this behavior for successful initialization, but it can break when using a volume to mount the data dir from another long-running file system (whether on the host or another container): when a the container is started on an empty volume, the mysql user may have write permissions there (because the volume is writable to anyone or because the volume was previously used to host another mysql database and the admin just `rm -rf $vol/*` it). If that is the case, then the first call to `mysqld --help` will create the innodb log files and then initialization will fail.

ltangvald:
This is a known issue with MySQL 5.7.6 and later, and is being looked into:
http://bugs.mysql.com/bug.php?id=75995. 
We have a temporary workaround in the mysql/mysql-server:5.7 image:
https://github.com/mysql/mysql-docker/blob/mysql-server/5.7/docker-entrypoint.sh
Note that the whole if-file-exists check can probably be replaced with a simple rm -f

ltangvald:
There's a better workaround for this; running mysqld --verbose --help with the --innodb-read-only option prevents it from creating the files. Could maybe make a PR for this, as it's not as ugly :)

tianon:
:+1: :metal:

It's not necessary in our latest version though, is it?  Are we still creating those files prematurely?

ltangvald:
[...]

PR #78, opened 2015-06-12
Prevent mysqld --verbose --help from writing to empty datadir

ltangvald:
From 5.7.6 and on, running mysqld --verbose --help will fully initialize innodb. If datadir is empty and writeable (with the default setup it's not writeable) this will create certain files there, causing the following initialization step to fail.

Using --innodb-read-only prevents this from happening, which should fix issue #69.

### DIFF
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -17,7 +17,7 @@ fi
 
 if [ "$1" = 'mysqld' ]; then
 	# Get config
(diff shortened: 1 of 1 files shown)


yosifkit:
Change LGTM.

On a side note, looks like the test are failing but that is unrelated to this change.

tianon:
LGTM, tested locally

I'm investigating the Travis issue, but it appears to somehow be related to heavy disk activity during the installation of the `mysql` package. :confused:

PR #80, opened 2015-06-17
load time zones into mysql

phemmer:
This enables time-zone support in mysql.

https://dev.mysql.com/doc/refman/5.7/en/time-zone-support.html

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -49,6 +49,8 @@ if [ "$1" = 'mysqld' ]; then
 			exit 1
 		fi
 
+		mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql --protocol=socket -uroot mysql
+
 		# These statements _must_ be on individual lines, and _must_ end with
 		# semicolons (no line breaks or comments are permitted).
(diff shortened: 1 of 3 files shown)


yosifkit:
LGTM

yosifkit:
ping @ltangvald

ltangvald:
I'm not familiar with the bug in question, but shouldn't the workaround be for both 5.5 and 5.6 (bug was first reported for 5.0)?

Otherwise it looks good.

phemmer:
> I'm not familiar with the bug in question, but shouldn't the workaround be for both 5.5 and 5.6 (bug was first reported for 5.0)?

I manually tested it on each of the versions, and 5.6 is the only one currently affected.

ltangvald:
LGTM.
Guess the weirdness of the 5.5-image sometimes works in our favor :P

tianon:
Can you update this to include the `--protocol` change from #84?  If not, that's alright -- I can do it post merge instead if you'd prefer. :+1:

phemmer:
@tianon Done.

tianon:
Thanks! :+1:

Issue #77, opened 2015-06-11
Travis mysql:5.7 runs timing out during build, after "debconf: falling back to frontend: Teletype"


edmorley:
eg:
https://travis-ci.org/docker-library/mysql/jobs/66373323

```bash
...
Setting up mysql-community-server (5.7.7-rc-1debian7) ...
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (This frontend requires a controlling tty.)
debconf: falling back to frontend: Teletype

No output has been received in the last ... minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

tianon:
This is since fixed. :smile: :+1:

Issue #76, opened 2015-06-11
Travis runs failing in mysql-basics with "mysqld failed to accept connetions in a reasonable amount of time!"


edmorley:
eg:
https://travis-ci.org/docker-library/mysql/jobs/66373324
```bash
Successfully built 6a18d5aa6546

The command "docker build -t "$image" ." exited with 0.

24.05s$ ~/official-images/test/run.sh "$image"
testing mysql:5.6
	'utc' [1/4]...passed
	'cve-2014--shellshock' [2/4]...passed
	'no-root-password' [3/4]...passed
	'mysql-basics' [4/4]...mysqld failed to accept connetions in a reasonable amount of time!
time="2015-06-11T12:33:59Z" level=fatal msg="Error response from daemon: Cannot start container 941844c4923e471eb7cfc25015bffbd6ef6064200f6ce6f631b9a41ac6bbfed8: Cannot link to a non running container: /mysql-container-22962-26882 AS /furious_albattani/mysql" 
failed

The command "~/official-images/test/run.sh "$image"" exited with 1.
```

Additionally, the mysql:5.7 runs are not even building, will file a separate issue for that.

edmorley:
Note: the above example run was with my PR changes reverted for mysql 5.6 - ie: the equivalent of master, so it should have been green.

Looking at the Travis, the last build on master was green - but that was 8 days ago, so presume either something has changed in the https://github.com/docker-library/official-images repo, or on Travis' experimental trusty stack (this repo uses `dist: trusty`).

tianon:
This is since fixed. :smile: :+1:

PR #84, opened 2015-07-03
Specify protocol for client during init

### LINKED ISSUE #83: mysql-first-time fails to run when custom cnf configuration has client protocol=TCP
mysql-first-time fails to run when custom cnf configuration has client protocol=TCP

1. Create custom MySQL configuration file with the following content -

```
[client]
protocol=tcp
```

2. Start mysql docker using the instructions from https://github.com/docker-library/docs/tree/master/mysql#user-content-using-a-custom-mysql-configuration-file

EXPECTING: mysql-first-time script runs

ACTUAL: fails with the following error -

ERROR 2003 (HY000): Can't connect to MySQL server on 'localhost' (111)

See line 81 https://github.com/docker-library/mysql/blob/master/5.6/docker-entrypoint.sh.  Shouldn't the protocol be forced to something other than TCP since mysqld was started with skip-networking?

ltangvald:
Image init can fail if user specifies a custom config with a changed socket setting (e.g. setting it to tcp) as mentioned in issue #83.
There's also a related problem in that the 5.5 image stores the default settings in a folder that will be overriden by the suggested method for providing a custom config (https://github.com/docker-library/docs/tree/master/mysql#user-content-using-a-custom-mysql-configuration-file).

Fixes #83

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -1,5 +1,4 @@
 FROM debian:wheezy
-
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN groupadd -r mysql && useradd -r -g mysql mysql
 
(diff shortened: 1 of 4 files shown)


edmorley:
Note the Travis failures are pre-existing - see #76 and #77.

tianon:
LGTM :+1:

PR #86, opened 2015-07-11
Allow creation of multiple databases

victorhahncastell:
See amended documentation at https://github.com/victorhahncastell/docker-docs/tree/master/mysql. Matching pull request will go to docker-library/docs

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -67,17 +67,83 @@ if [ "$1" = 'mysqld' ]; then
 			DROP DATABASE IF EXISTS test ;
 		EOSQL
 
-		if [ "$MYSQL_DATABASE" ]; then
-			echo "CREATE DATABASE IF NOT EXISTS \`$MYSQL_DATABASE\` ;" >> "$tempSqlFile"
-		fi
+		# Create databases if specified on command line
+                i=0
+                while :
+                do
+                        if [ "$i" -eq "0" ]; then             # support "legacy" option before multiple database support
+                                this_varname=MYSQL_DATABASE
+                        else
+                                this_varname=MYSQL_DATABASE$i
+                        fi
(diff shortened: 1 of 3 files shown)


yosifkit:
This seems a bit complex for an official image and I would rather recommend that a user just write a startup script to creates all the users, databases, and permissions that they need, but I'll defer to @ltangvald since he is from upstream and helps manage these images.

tianon:
I agree, this is incredibly complex (both the implementation and the user
experience).

victorhahncastell:
Thanks for your feedback! I agree the implementation is a bit complex--it probably shows that I don't write shell scripts every day. I don't quite see though how the user experience is overly complicated, took me three sentences to amend the documentation.

ltangvald:
I agree the implementation is too complex. User experience seems ok, though ideally we'd have something that isn't any different to use whether you want 1 or more databases (e.g. you supply a delimited list to a single variable). I'll see if I can pick some brains about it (though almost everyone is on vacation at the moment).
If we can't find a clean way to implement it, finding a good solution to issue #24 might be a decent compromise.

ltangvald:
Is this now adequately solved by the support for running arbitrary init scripts during database creation?

asymmetric:
I think the answer to @ltangvald's question is yes, this and #18 should be closed since we can drop scripts in `/docker-entrypoint-initdb.d`.

victorhahncastell:
Guess I should comment on this as this is my PR -- though I don't really have any opinion on this as my work currently does not involve this project any more. Feel free to close :-)

yosifkit:
:+1:

mrts:
And here's a sample script and README with instructions for creating any number of databases during initialization by mounting a volume into `/docker-entrypoint-initdb.d`:
https://github.com/mrts/docker-posgresql-multiple-databases

PR #87, opened 2015-07-23
Add custom initialization hooks during database initialization

marclennox:


yosifkit:
LGTM, pending replication to the other versions.  Looks just like the one on [postgres](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L65-L69) :+1:, but the user can just use the socket connection (`mysql --protocol=socket -uroot -ppass`).

I think it would be useful to also allow `.sql` files in there and just do the mysql command for them.  Otherwise, we will need to document that their scripts would need to use the user and password that they set (or root and its password), which are all available in the env.

This is a fix for #24.

Ping @ltangvald and @tianon

tianon:
I agree, I think something like this would be somewhat more flexible (and is what I was considering for PostgreSQL, but then found out that `--single` isn't actually recommended):

```bash
		echo
		for f in /docker-entrypoint-initdb.d/*; do
			case "$f" in
				*.sh)  echo "$0: running $f"; . "$f" ;;
				*.sql) echo "$0: running $f"; gosu postgres postgres --single -jE < "$f" && echo ;;
				*)     echo "$0: ignoring $f" ;;
			esac
			echo
		done
```

marclennox:
I like the idea of supporting both sql and sh scripts for both pg and mysql.  For now I've created a mysql image (https://registry.hub.docker.com/u/tenstartups/mysql/dockerfile/) that has my proposed change (which indeed I copied directly from the postgres image).  Anyway, it works like a charm.

tianon:
https://github.com/docker-library/postgres/pull/75 is my attempt to bring the Postgres image into line with MySQL's init process (and finally axe the practice of using `postgres --single` which upstream discourages), also adding support for `*.sql` officially.

ltangvald:
Looks good to me too. Should also be mentioned in the documentation, though.

PR #89, opened 2015-07-30
Skip reverse DNS lookups

### LINKED ISSUE #52: Connection timeout with nodejs app occasionally
I am using docker-compose to connect a nodejs app to a mysql database with the following `.yml` file:
```yml
db:
  image: mysql
  ports:
    - "3306:3306"
  environment:
    - MYSQL_ROOT_PASSWORD=password
    - MYSQL_DATABASE=sane
server:
  image: artificial/docker-sails:iojs-full-stable-pm2
  command: sails lift --verbose
  volumes:
    - server/:/server
  ports:
    - "1337:1337"
  links:
    - db
```

Often it works, but often I get a timeout and a warning `'2015-03-04 14:40:37 1 [Warning] IP address '172.17.0.15' could not be resolved: Temporary failure in name resolution'` I am not entirely sure if the issue is here, or in the app itself, since if I switch to the postgresql I have no issues as well.

Still trying to nail down where exactly the issue arises, this is more to keep track of it in case it is an issue with the image being slow or something.

yosifkit:
fixes #52

### DIFF
--- a/5.6/Dockerfile
+++ b/5.6/Dockerfile
@@ -30,7 +30,10 @@ RUN { \
 	&& rm -rf /var/lib/mysql && mkdir -p /var/lib/mysql
 
 # comment out a few problematic configuration values
-RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf
+# don't reverse lookup hostnames, they are usually another container
+RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf \
+	&& echo 'skip-host-cache\nskip-name-resolve' | awk '{ print } $1 == "[mysqld]" && c == 0 { c = 1; system("cat") }' /etc/mysql/my.cnf > /tmp/my.cnf \
+	&& mv /tmp/my.cnf /etc/mysql/my.cnf
 
 VOLUME /var/lib/mysql
 
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -40,6 +40,8 @@ ENV PATH $PATH:/usr/local/mysql/bin:/usr/local/mysql/scripts
 RUN mkdir -p /etc/mysql/conf.d \
 	&& { \
 		echo '[mysqld]'; \
+		echo 'skip-host-cache'; \
+		echo 'skip-name-resolve'; \
 		echo 'user = mysql'; \
 		echo 'datadir = /var/lib/mysql'; \
 		echo '!includedir /etc/mysql/conf.d/'; \
--- a/5.7/Dockerfile
+++ b/5.7/Dockerfile
@@ -30,7 +30,10 @@ RUN { \
 	&& rm -rf /var/lib/mysql && mkdir -p /var/lib/mysql
 
 # comment out a few problematic configuration values
-RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf
+# don't reverse lookup hostnames, they are usually another container
+RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf \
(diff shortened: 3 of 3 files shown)


yosifkit:
```console
$ ./generate-stackbrew-library.sh | sed -r 's/docker-library/infosiftr/' > mysql
$ ./bashbrew/bashbrew.sh build ../mysql/mysql
Cloning mysql (git://github.com/infosiftr/mysql) ...
Processing mysql:5.5.45 ...
Processing mysql:5.5 ...
Processing mysql:5.6.26 ...
Processing mysql:5.6 ...
Processing mysql:5 ...
Processing mysql:latest ...
Processing mysql:5.7.7-rc ...
Processing mysql:5.7.7 ...
Processing mysql:5.7 ...
$ ./bashbrew/bashbrew.sh list --uniq ../mysql/mysql | xargs ./test/run.sh
testing mysql:5.5.45
	'utc' [1/4]...passed
	'cve-2014--shellshock' [2/4]...passed
	'no-hard-coded-passwords' [3/4]...passed
	'mysql-basics' [4/4]...passed
testing mysql:5.6.26
	'utc' [1/4]...passed
	'cve-2014--shellshock' [2/4]...passed
	'no-hard-coded-passwords' [3/4]...passed
	'mysql-basics' [4/4]...passed
testing mysql:5.7.7-rc
	'utc' [1/4]...passed
	'cve-2014--shellshock' [2/4]...passed
	'no-hard-coded-passwords' [3/4]...passed
	'mysql-basics' [4/4]...mysqld failed to accept connections in a reasonable amount of time!
ERROR 2003 (HY000): Can't connect to MySQL server on 'mysql' (111)
failed
```

5.6/Dockerfile
```diff
@@ -30,7 +30,10 @@ RUN { \
 	&& rm -rf /var/lib/mysql && mkdir -p /var/lib/mysql
 
 # comment out a few problematic configuration values
-RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf
+# don't reverse lookup hostnames, they are usually another container
+RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf \
+	&& echo 'skip-host-cache\nskip-name-resolve' | awk '{ print } /^\[mysqld\]$/ && c == 0 { c = 1; system("cat") }' /etc/mysql/my.cnf > /tmp/my.cnf \
```
tianon: This regex condition could instead be `$1 == "[mysqld]"`.

tianon:
I'm definitely a fan of skipping these by default for the image, especially since so many users will be behind at least Docker's NAT by default, so incoming connections won't resolve and we're just wasting precious time.

ping @ltangvald; what're your thoughts here?

yosifkit:
This leaves the `my.cnf` having this diff from it's previous state:
```diff
 [mysqld]
+skip-host-cache
+skip-name-resolve
 user		= mysql
 pid-file	= /var/run/mysqld/mysqld.pid
 socket		= /var/run/mysqld/mysqld.sock
```

shofetim:
:+1:

tianon:
LGTM

joshunger:
I believe this breaks permissions/grants in certain cases?  Can this be a setting?  Or, how do you override "skip-name-resolve" in a file in the conf.d directory?

tianon:
Sorry for the delay, @joshunger; I would _imagine_ something like `skip-name-resolve = off` or `skip-name-resolve = false` should work (given that the config is supposed to be key-value pairs), but I haven't tested to verify that.  If that doesn't work, the only other thing I can think of (besides overwriting this provided file completely) is to do something like `RUN sed -i '/skip-name-resolve/d' ...` to remove the line.

PR #90, opened 2015-08-04
Add support for "/docker-entrypoint-initdb.d" similar to PostgreSQL's implementation

### LINKED ISSUE #24: Provide mechanism to extend mysql image with additional SQL init scripts

tianon:
This also cleans up some minor unnecessary bits in `docker-entrypoint.sh` so that it's more consistent.

Compare to https://github.com/docker-library/postgres/blob/a82c28e1c407ef5ddfc2a6014dac87bcc4955a26/docker-entrypoint.sh#L49-L97

Fixes #24
Closes #87 (implements what was additionally discussed there)

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -1,17 +1,6 @@
 #!/bin/bash
 set -e
 
-get_option () {
-	local section=$1
-	local option=$2
-	local default=$3
-	# my_print_defaults can output duplicates, if an option exists both globally and in
-	# a custom config file. We pick the last occurence, which is from the custom config.
-	ret=$(my_print_defaults $section | grep '^--'${option}'=' | cut -d= -f2- | tail -n1)
-	[ -z $ret ] && ret=$default
-	echo $ret
-}
-
 # if command starts with an option, prepend mysqld
 if [ "${1:0:1}" = '-' ]; then
 	set -- mysqld "$@"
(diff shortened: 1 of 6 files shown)


tianon:
cc @ltangvald @yosifkit

5.5/docker-entrypoint.sh
```diff
@@ -38,25 +24,25 @@ if [ "$1" = 'mysqld' ]; then
 		mysql_install_db --user=mysql --datadir="$DATADIR" --rpm --basedir=/usr/local/mysql
 		echo 'Finished mysql_install_db'
 
-		mysqld --user=mysql --datadir="$DATADIR" --skip-networking --basedir=/usr/local/mysql --pid-file="$PIDFILE" &
-		for i in $(seq 30 -1 0); do
-			[ -S "$SOCKET" ] && break
+		mysqld --user=mysql --datadir="$DATADIR" --skip-networking --basedir=/usr/local/mysql &
+		pid="$!"
```
tianon: By using `$!` instead of parsing the `mysqld` config, we can guarantee that the process is a child of our shell and thus we can `wait` on it during shutdown (see below) instead of looping while we wait for it to die.

yosifkit:
LGTM

funkyfuture:
Are there docs on this functionality?

yosifkit:
Looks like we forgot to do that. it should be fairly similar to the one for postgres: https://hub.docker.com/_/postgres/#how-to-extend-this-image

ltangvald:
Very nice! The upcoming http://dev.mysql.com/doc/refman/5.7/en/shutdown.html might help clean up the entrypoint script further.

thaJeztah:
Thanks @ltangvald! And thanks @tianon I think this could make many people happy

funkyfuture:
here's one: :dancer:

tianon:
@ltangvald oh interesting -- I'll bet we could just use `mysqladmin` directly right now, couldn't we?

ltangvald:
No, unfortunately I got a bit too excited about it :)
mysqladmin shutdown will send a shutdown signal, but won't wait for a complete server shutdown.
So the SHUTDOWN statement won't be very useful as it is now. We'd need to put everything into a single init-file with the syntax constraints that imply.

PR #91, opened 2015-08-09
Upgrade base image to debian:jessie

sherter:
It is likely that this image is used together with [php/apache](https://github.com/docker-library/php) or [nginx](https://github.com/nginxinc/docker-nginx). Having the same base image for all those services saves the user some space. And using the latest stable debian release is a good idea anyway.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:wheezy
+FROM debian:jessie
 
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN groupadd -r mysql && useradd -r -g mysql mysql
--- a/5.6/Dockerfile
+++ b/5.6/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:wheezy
+FROM debian:jessie
 
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN groupadd -r mysql && useradd -r -g mysql mysql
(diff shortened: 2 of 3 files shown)


yosifkit:
LGTM

tianon:
I'm on the fence with this one.  Just swapping the base like this _might_ work, but could also cause issues with the following line later down in the `Dockerfile`:
```Dockerfile
RUN echo "deb http://repo.mysql.com/apt/debian/ wheezy mysql-${MYSQL_MAJOR}" > /etc/apt/sources.list.d/mysql.list
```

However, it _does_ appear that upstream has a `jessie` dist, so that's worth trying: https://repo.mysql.com/apt/debian/dists/jessie/

tianon:
cc @ltangvald

sherter:
@tianon changed dist in 5.6 and 5.7 to jessie and squashed commits. All tests pass locally.

ltangvald:
Switching to our jessie package should work fine, but I'll look into it a bit to make sure there aren't any differences we need to be aware of.

sherter:
ping @ltangvald

ltangvald:
Sorry about the late reply :)

I haven't found any changes that would be significant for Docker, so I think this should work fine.

sherter:
Thanks :+1: 
I guess this can be merged then @tianon @yosifkit.

yosifkit:
LGTM (the build fails are just travis issues like `/entrypoint.sh: line 85: kill: (61) - Permission denied` and the input hang on 5.7)

tianon:
Nice, LGTM :+1:

Issue #92, opened 2015-08-18
Bug on MySQL container startup when using MYSQL_ALLOW_EMPTY_PASSWORD


vincentchalamon:
When using `MYSQL_ALLOW_EMPTY_PASSWORD` environment variable in my _docker-compose.yml_ configuration, MySQL is unable to restart because of following error:
> MySQL init process in progress...
> …
> Warning: Unable to load '/usr/share/zoneinfo/iso3166.tab' as time zone. Skipping it.
> Warning: Unable to load '/usr/share/zoneinfo/zone.tab' as time zone. Skipping it.
> 2015-08-18 15:51:19 60 [Warning] 'proxies_priv' entry '@ root@default' ignored in --skip-name-resolve mode.
> Enter password: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)

I don't have the problem when using a password. Here is my _docker-compose.yml_ configuration:
```yml
web:
    build: .
    volumes:
        - .:/var/www
    net: "host"
    tty: true

db:
    image: mysql
    command: mysqld --port 3386
    net: "host"
    environment:
        MYSQL_DATABASE: api
        MYSQL_USER: root
        MYSQL_ALLOW_EMPTY_PASSWORD: yes
```

I didn't have this problem a few weeks ago, I just reset my docker environment.

yosifkit:
Whoops, looks like we just need to adjust this line: https://github.com/docker-library/mysql/blob/master/5.6/docker-entrypoint.sh#L58 to check if the password is required.

PR #93, opened 2015-08-20
Fix MYSQL_ALLOW_EMPTY_PASSWORD check

vincentchalamon:
Fix https://github.com/docker-library/mysql/issues/92

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -48,14 +48,17 @@ if [ "$1" = 'mysqld' ]; then
 			-- What's done in this file shouldn't be replicated
 			--  or products like mysql-fabric won't work
 			SET @@SESSION.SQL_LOG_BIN=0;
-			
+
 			DELETE FROM mysql.user ;
 			CREATE USER 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
 			GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION ;
 			DROP DATABASE IF EXISTS test ;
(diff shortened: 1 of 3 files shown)


MLKiiwy:
:+1:

yosifkit:
LGTM

5.6/docker-entrypoint.sh
```diff
@@ -48,14 +48,17 @@ if [ "$1" = 'mysqld' ]; then
...
+
 			DELETE FROM mysql.user ;
 			CREATE USER 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
 			GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION ;
 			DROP DATABASE IF EXISTS test ;
 			FLUSH PRIVILEGES ;
 		EOSQL
-		mysql+=( -p"${MYSQL_ROOT_PASSWORD}" )
+
+		if [ -z "$MYSQL_ALLOW_EMPTY_PASSWORD" ]; then
```
tianon: This needs to just be `-z "$MYSQL_ROOT_PASSWORD"` or else we'll fail on things like `-e MYSQL_ALLOW_EMPTY_PASSWORD=1 -e MYSQL_ROOT_PASSWORD=something` (which currently works).

We already bail above on `MYSQL_ALLOW_EMPTY_PASSWORD`, so at this point if `MYSQL_ROOT_PASSWORD` is empty, we have an empty password.

guillaumevoisin:
:+1:

ltangvald:
Teach me to look at the issues/prs before doing work. Was about to submit this exact PR, so LGTM :)

PR #109, opened 2015-10-21
Fix some quotes, simplify update script

yosifkit:
Also, syncing in the `MYSQL_INITDB_SKIP_TZINFO` env variable and use "$@" instead of "mysqld" so that we use passed in options too.


tianon:
LGTM

ltangvald:
LGTM

PR #114, opened 2015-10-27
Removed --innodb-read-only workaround for fixed server bug

ltangvald:
The option was added as a workaround for a server bug that caused file locks in datadir when
mysqld --verbose --help is run. The bug has been fixed, so the workaround is no longer needed.

MySQL bug: http://bugs.mysql.com/bug.php?id=75995


tianon:
Nice! :heart: 

LGTM

Issue #47, opened 2015-01-31
Container does not catch signals and exit (Ctrl+C)


msumme:
If I start a database like this:

```
/usr/bin/docker run --name database -e MYSQL_ROOT_PASSWORD=whatever mysql:5.6
```
The container does not respond to Ctrl+c like most containers do - it simply will not exit until I manually kill the process from a different terminal.

Even if I add "-it" options, it is still the same.

This is on docker 1.4.1.

Other containers exhibit the expected behavior on the same docker version.

yosifkit:
This is a limitation of docker and pid 1.  If the process doesn't implement a signal handler then the SIGINT is just lost because normally you would not want to kill your pid 1 which is your init system.  Redis has the same problem.

msumme:
I figured that out and found a workaround.

It is possible to run mysqld in a wrapper that properly handles signals so that the instance can be monitored externally and shut down with normal signal handling.  

This can be accomplished by changing the last line from `exec "$@` to:
```
$@ &
PID=`jobs -p`

trap "kill -SIGQUIT $PID" INT
wait
```

I'm sure something more sophisticated could be accomplished, but it would be nice to be able to monitor your msyql container w/ systemd and other init systems.

md5:
Systemd and other init systems don't use SIGINT, they use SIGTERM, so this
isn't a problem in those situations.​

msumme:
Okay - well, I'll have to check again - but I couldn't seem to get it to work with systemd until I changed the script.

vaab:
Here's how to use @msumme fix, without changing the entrypoint. This will work with a pristine image. Please, don't judge me.

```
docker run -ti \
    --entrypoint /bin/bash mysql -c '                                                                                          
    exec() {                                                                                                                     
        "$@" &                                                                                                                   
        pid="$!"                                                                                                                 
        trap "kill -SIGQUIT $pid" INT                                                                                            
        wait                                                                                                                     
    }                                                                                                                            
    export -f exec                                                                                                               
    /entrypoint.sh '

**Comment by jeremyVignelles** [comment]
Hi,
Same issue here, I was looking for a clean solution when I came across this thread.

I tried @vaab solution but I couldn't make it work...
I've ended up doing my own Dockerfile, like this one:

```
FROM mysql:latest

ENV MYSQL_ROOT_PASSWORD password
ENV MYSQL_DATABASE databasename
ENV MYSQL_USER user
ENV MYSQL_PASSWORD password

[...]

PR #117, opened 2015-11-03
Exit gracefully on ctrl-C

jeremyVignelles:
When mysql is run in interactive mode, with the following command line, I can't stop it by hitting `ctrl-c`.

```
docker run --rm -it -e MYSQL_ROOT_PASSWORD=a mysql:latest
```
The two ways I found to stop it were:

* `docker kill`
* Creating my own Dockerfile that wraps the entrypoint in a bash script that handles SIGINT as can be seen in my fix.

I think this fix can be merged in the official release. Please let me know if this has unwanted side-effects...

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -100,4 +100,12 @@ if [ "$1" = 'mysqld' ]; then
 	chown -R mysql:mysql "$DATADIR"
 fi
 
-exec "$@"
+"$@" &
+pid="$!"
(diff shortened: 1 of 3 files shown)


tianon:
This is a duplicate of #47.  I think the biggest unwanted side effect here is the stray `bash` process this leaves behind that we don't have otherwise.  Additionally, given that MySQL _does_ respond appropriately to `SIGTERM` (which is what `docker stop`/`docker restart` uses), and explicitly does _not_ respond to `SIGINT`, I'm inclined to defer to upstream's judgement here -- they might have a reason that they've chosen not to listen for `SIGINT` explicitly, and I'd hate to deviate if we don't have a really solid reason for it.

PR #119, opened 2015-11-23
Added security options

ltangvald:
* Added option for randomly generated root password
* Added option to immediately expire root password
---
Installs pwgen package
MYSQL_RANDOM_ROOT_PASSWORD:
Uses pwgen to generate a password for the server's root user, and outputs it to the containers stdout

MYSQL_ONETIME_PASSWORD:
Sets root user as expired once init is complete, forcing a password change on first login (5.6+)

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -11,12 +11,11 @@ if [ "$1" = 'mysqld' ]; then
 	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
 
 	if [ ! -d "$DATADIR/mysql" ]; then
-		if [ -z "$MYSQL_ROOT_PASSWORD" -a -z "$MYSQL_ALLOW_EMPTY_PASSWORD" ]; then
-			echo >&2 'error: database is uninitialized and MYSQL_ROOT_PASSWORD not set'
-			echo >&2 '  Did you forget to add -e MYSQL_ROOT_PASSWORD=... ?'
-			exit 1
-		fi
-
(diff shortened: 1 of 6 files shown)


5.6/docker-entrypoint.sh
```diff
@@ -11,12 +11,11 @@ if [ "$1" = 'mysqld' ]; then
...
-			echo >&2 'error: database is uninitialized and MYSQL_ROOT_PASSWORD not set'
-			echo >&2 '  Did you forget to add -e MYSQL_ROOT_PASSWORD=... ?'
-			exit 1
-		fi
-
+		if [ -z "$MYSQL_ROOT_PASSWORD" -a -z "$MYSQL_ALLOW_EMPTY_PASSWORD" -a -z "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
+                        echo >&2 'error: database is uninitialized and password option is not specified '
+                        echo >&2 '  You need to specify one of MYSQL_ROOT_PASSWORD, MYSQL_ALLOW_EMPTY_PASSWORD and MYSQL_RANDOM_ROOT_PASSWORD'
+                        exit 1
+                fi
```
tianon: Whitespace here seems a little off -- is this tabs, or spaces?

tianon:
Very nice additions IMO :+1: :metal:

5.6/docker-entrypoint.sh
ltangvald: Yeah, looks like a mix. Will fix :)

ltangvald:
Should be fixed now :)

yosifkit:
LGTM

tianon:
LGTM

PR #120, opened 2015-11-23
Made charset and collate customizable, with utf8 as default.

OmarIthawi:
I was a little bit frustrated that this docker instance creates a `latin1` database by default. I needed a way to customize it.

TODO:
* [ ]  Update the documentation, but would like to get feedback on this PR if possible.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -10,6 +10,10 @@ if [ "$1" = 'mysqld' ]; then
 	# Get config
 	DATADIR="$("$@" --verbose --help 2>/dev/null | awk '$1 == "datadir" { print $2; exit }')"
 
+	# Set the default charset to UTF8 if not set.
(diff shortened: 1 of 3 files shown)


Zophar78:
Exactly what i was looking for!!! +1
But it has already done a lot of noise here: https://github.com/docker-library/mysql/pull/14

OmarIthawi:
I've used it and liked it. But looks like the community have other requirements as discussed in #14.

I'm closing this for now.

ltangvald:
You might add your thoughts there. Also take a look at the ability to pass any server option to the image, and see if that would help in your case.

OmarIthawi:
@ltangvald My requirements were very simple. I needed the default database to be UTF-8 and got that with this PR. It's up to the repo owner to decide whether to take the PR or not :).

OmarIthawi:
@ltangvald I saw the efforts on the server locale. But I'm not sure that I would need it. Thanks :)

PR #143, opened 2016-02-16
Add MYSQL_CUSTOM_DIR variable

ruslanloman:
Add MYSQL_CUSTOM_DIR variable to be able to create directories and change owners during launch of container

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -107,6 +107,13 @@ if [ "$1" = 'mysqld' ]; then
 		echo
 	fi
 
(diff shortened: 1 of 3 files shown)


ltangvald:
What is the use case for this?
I think this would be better solved by using the existing /docker-entrypoint-initdb.d system for running init  scripts (i.e. adding an .sh script that creates the directories and sets ownership), though a current limitation is that it only runs during database creation.

ruslanloman:
Thanks for your advice with  /docker-entrypoint-initdb.d, this approach is working for me.

Issue #136, opened 2016-01-28
set -eo pipefail addition to docker-entrypoint.sh in ed198ce prevents mysqld from booting


hhowe29:
We have a mysql docker image that we could build and boot fine before this commit from 1/20. After this change, mysqld never gets going. No logging is omitted. We traced this to the fact that our my.cnf file contained a log-bin setting:

log-bin=foo-bin
 
We were able to workaround the problem in two different ways: 
1: revert to the previous entrypoint script. ADD it at from the Dockerfile and overwrite the newer script
or
2: remove the log-bin setting from our my.cnf config. 

Posting this in case someone else runs into a similar issue. Our problem seems related to this: https://bugs.mysql.com/bug.php?id=78957

**Details:**
docker version 1.9.1
fig version 1.0.1
uname -a = Linux vagrant-ubuntu-trusty-64 3.13.0-74-generic #118-Ubuntu SMP Thu Dec 17 22:52:10 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

**./mysql/Dockerfile**:
```
FROM mysql:5.6.28
run apt-get update && apt-get install -qq -y nano
ADD my.cnf /etc/my.cnf
```

**./mysql/my.cnf:**
```
[mysqld]
bind-address=0.0.0.0
log-bin=foo-bin
console=1
general_log=1
general_log_file=/var/lib/mysql/mysql-general.log
slow_query_log=1
slow_query_log_file=/var/lib/mysql/mysql-slow.log
log_error=/var/lib/mysql/mysql-error.log
collation-server=utf8_unicode_ci
character-set-server=utf8
binlog_format=ROW
server_id=999
transaction-isolation=READ-COMMITTED
sync_binlog=1
sync_frm=1
log_output=TABLE,FILE
```
**./fig file**
```
db:
  build: mysql
  ports:
   - 3306:3306
  environment:
   MYSQL_ROOT_PASSWORD: supersecret
   MYSQL_DATABASE: fst
   MYSQL_USER: fst
   MYSQL_PASSWORD: ubersecret
```

**commands:**
```
fig rm --force db
fig build db
fig up db
```

ltangvald:
Thanks for the report (and trigger info). We'll take a closer look at it :)

yosifkit:
[...]

PR #144, opened 2016-02-17
Workaround for MySQL bug #78957, which also affects 5.5 and 5.6

ltangvald:
Adding --log-bin-index will change where the index file is written,
but has no effect if --log-bin has not been enabled

--
This should fix issue #136


tianon:
Nice, LGTM! :+1:

PR #147, opened 2016-02-24
Skip setup if they want to get --help

yosifkit:
This will allow users to see available flags with `docker run -it --rm mysql --verbose --help`.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -6,7 +6,18 @@ if [ "${1:0:1}" = '-' ]; then
 	set -- mysqld "$@"
 fi
 
-if [ "$1" = 'mysqld' ]; then
+# skip setup if they want an option that stops mysqld
+wantHelp=
+for arg; do
+	case "$arg" in
+		-'?'|--help|--print-defaults|-V|--version)
+			wantHelp=1
+			break
+			;;
+	esac
+done
+
+if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 	# Get config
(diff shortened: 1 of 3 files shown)


tianon:
@ltangvald what're your thoughts on this?  Right now it falls over pretty hard if we use `--help`...  (tries to initialize a database, etc)

Do you think there are other `mysqld` options we should add to the "bypass" list? (maybe start with this and keep our eyes open for more examples as we go?)

ltangvald:
Maybe --print-defaults?

ltangvald:
A concern is that this will quickly balloon out of control if we start adding checks for any options that cause the server to behave in a non-standard way. E.g. -? is a variant of --help and would also need to be included. Maybe better to either try to detect the failure early so the user doesn't have to wait for a timeout or have a specific, documented way of getting a readme-like output from the image?

yosifkit:
It could get longer, yes, but I think there is a limited set that cause the server to not actually run that we should catch, rather than trying to detect that a failure is a success. We would otherwise have to detect them in two spots, one after trying to get `DATADIR` and skip the rest of the setup, and the other would be when we start the server in the background, which could've failed for other reasons.

The set I have found is: `-'?'|--help|--print-defaults|-V|--version`.  The user will have to be aware that `--print-defaults` needs to be the first argument, but that is in the verbose help output and should not be something we need to explicitly test for.

yosifkit:
@ltangvald, does this seem like a solid solution?

ltangvald:
Gah, sorry. Yeah as long as we keep it limited it looks good :)

tianon:
LGTM :+1:

PR #148, opened 2016-02-29
Fix "gpg" usage to stop relying on deprecated and insecure behavior

tianon:
See also docker-library/official-images#1420.


yosifkit:
LGTM

PR #151, opened 2016-03-13
If n is not given, the command waits until all jobs known to the invoking shell have terminated.

rusxakep:
For better compatibility, wait for all child.


rusxakep:
In previous comment, I said about my problem in "ubuntu and bash 4.xx" environment.
This isn't correct. Sure, problem with TERM exit code in bash version from Dockerfile ("Debian 8")

yosifkit:
Sorry for the long tail on this issue.  I'm not entirely sure what problem that this is solving.  I would rather not change the wait, since changing it to all children could break current users experience if they had purposely started something in the background in one of their shell scripts in `/docker-entrypoint-initdb.d/`.

tianon:
I agree with @yosifkit -- I think this would be a backwards-incompatible change, and is better left to the individual `initdb` scripts to handle (either by not backgrounding or by invoking `wait` appropriately themselves).

Thanks! :heart:

PR #159, opened 2016-04-06
Increase the 30 seconds check 

shantanuo:
container fails to start when initializing the log file size takes more than 30 seconds to complete. Please increase this limit.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -51,7 +51,6 @@ RUN mkdir -p /etc/mysql/conf.d \
 		echo '!includedir /etc/mysql/conf.d/'; \
 	} > /etc/mysql/my.cnf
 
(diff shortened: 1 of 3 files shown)


yosifkit:
Perhaps we should just make it configurable through an environment variable passed in the docker run?  I feel like most systems start quickly enough that we would rather fail fast and then allow a user to adjust if their system needs a little more time.

What do you think @ltangvald?

ltangvald:
Yeah, passing it might be better. As you say it's usually enough, though I've also seen it fail on certain systems. Plus, 5.7 and 8.0 are slower to start up than the earlier versions.

yosifkit:
Apologies for the delay :bowing_man:

Closing in favor of https://github.com/docker-library/mysql/issues/646

PR #161, opened 2016-04-15
Allow arbitrary --user values

### LINKED ISSUE #153: mysql:5.5: can't chown on nfs mount
Running mysql:5.5 with `/var/lib/mysql` mounted on a nfs file system, I get this error:

    chown: changing ownership of '/var/lib/mysql/': Invalid argument

I edited the entrypoint script to see if I had write access in `/var/lib/mysql`, and I have, and files I create in the entrypoint script are created in the mounted directory. It's just the chown that fails, and at that point, the script fails.

It works fine on an ext4 file system on the same machine, and then the result of chown is `999:localadmin`

tianon:
Closes #153

See also docker-library/rabbitmq#60, docker-library/cassandra#48, docker-library/mongo#81, and docker-library/redis#48.

cc @ltangvald

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -3,6 +3,20 @@ FROM debian:jessie
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN groupadd -r mysql && useradd -r -g mysql mysql
 
+# add gosu for easy step-down from root
+ENV GOSU_VERSION 1.7
+RUN set -x \
+	&& apt-get update && apt-get install -y --no-install-recommends ca-certificates wget && rm -rf /var/lib/apt/lists/* \
+	&& wget -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$(dpkg --print-architecture)" \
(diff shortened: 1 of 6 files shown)


tianon:
According to https://www.mysql.com/support/eol-notice.html, we might technically be able to drop 5.5 as EOL as of today (although we're running it on Debian 7, which never was actually supported), but since it was pretty simple to make it work too, I just went for it.

ltangvald:
5.5 is actually pretty nice to have in that upstream doesn't actually provide Debian/Ubuntu packages for it, so I'd think the ease of use/setup of the Docker image is a pretty big plus (though I have no usage numbers to back this up with :P)

tianon:
Fair enough! :+1:

ltangvald:
Oh, and change generally looks good to me, though I have to admit I haven't looked closely at gosu :)

tianon:
It's conceptually really simple -- like `sudo`, but doesn't touch the TTY or stay resident, just `setuid`+`setgid`+`setgroups`+`exec`. :+1:

yosifkit:
I think the main difference is that the `initdb.d` scripts will no longer be run as root; not sure if that will break too many users.

yosifkit:
LGTM.  Should we merge?

ltangvald:
Yeah, I think we can.
As an aside, I forgot to consider when I made the comment about 5.5 that since it was only available on Debian6, it'll no longer be updated for any Debian/Ubuntu platform so keeping it as it is wouldn't be a good idea (issue #168).

PR #181, opened 2016-06-23
Fix start failure when MYSQL_PWD env variable is set

cdwertmann:


cdwertmann:
Initialisation of the database fails when the MYSQL_PWD env variable is set in the docker container

tianon:
I think it's probably prudent if we discuss this over in #184 before we implement anything, especially since it sounds like `MYSQL_PWD` isn't the only variable that can have an effect here, and the others (such as `MYSQL_HOST`) aren't necessarily as easy to work around.

yosifkit:
Fixed by https://github.com/docker-library/docs/pull/1105

PR #182, opened 2016-06-23
Remove volumes

dagguh:
Docker doesn't allow removing a volume from an existing image.


dagguh:
Oops, I wanted to target my fork with this.

PR #191, opened 2016-07-12
Add $MYDATA env variable (like $PGDATA in postgres official image) and set working dir

mouhamed:


tianon:
The difference between this and `PGDATA` is that `PGDATA` is an environment variable supported and used directly by PostgreSQL itself, not something created specifically just for the Dockerization.

You _should_ be able to accomplish this using something like `--datadir=/somewhere/else` as your command (ie, `docker run some-image --datadir=/somewhere/else` or `CMD ["--datadir=/somewhere/else"]` in a `Dockerfile` that's `FROM` this image).

Ulexus:
@tianon Unfortunately, while running the container on an already-created datastore will work by passing `--datadir`, the initialization of the database will _NOT_ work, because there is no way to override the value of DATADIR right now.  

Ref #207

yosifkit:
This pull request intent is easily accomplish by adding `--datadir` to your command as stated by @tianon above and detailed over [here](https://github.com/docker-library/mysql/pull/208#issuecomment-245035207).  If this is something that should be specifically added to the documentation on the Docker Hub, then it can be added over in the docs repo under [mysql/content.md](https://github.com/docker-library/docs/blob/c4989f32d2fe2bf8404026ba7d92fd7bfc82aa62/mysql/content.md).

Issue #192, opened 2016-07-21
Allow multiple volumes to be mounted as init scripts for mysql


mtzmontiel:
Currently this image loads only the first level on path: 

```
 /docker-entrypoint-initdb.d/
```

So if some wanted to use multiple volumes mounted under that only the lowest level would be executed.

I want to use volumes or to order init scripts in a more meaningful manner:

```
/docker-entrypoint-initdb.d/00-create-schemas.sql
/docker-entrypoint-initdb.d/10-myschema-seed/10-schema-1.sql
/docker-entrypoint-initdb.d/20-myschema-seed/10-schema-2.sql
/docker-entrypoint-initdb.d/30-myschema-seed/10-schema-3.sql
```

tianon:
Closing for the same reasons as the associated PR: https://github.com/docker-library/mysql/pull/193

tianon:
(Suggested workaround is to drop a short `.sh` file instead which sources additional directories.)

PR #193, opened 2016-07-21
Allow deep nesting on init path.

mtzmontiel:
FIXES: https://github.com/docker-library/mysql/issues/192

---

This change is [Reviewable](https://reviewable.io/reviews/docker-library/mysql/193)

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -104,7 +104,7 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 		fi
 
 		echo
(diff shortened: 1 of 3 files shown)


ltangvald:
LGTM

@yosifkit

yosifkit:
There is a related discussion on https://github.com/docker-library/postgres/issues/179 and I echo my concern for breaking changes.

ltangvald:
Yeah, I see the issue. An alternative (though slightly more work for the user) is to add a .sh file to the initdb.d directory that includes the rest of them, since that will already work

Issue #197, opened 2016-08-02
Latest 5.7 container (5.7.14) seems to stopped working with custom user IDs anymore


ffdybuster:
Having `/data/mysql` owned by 1000:1000, when I try to run:

```
docker run -d --name mysql -u 1000:1000 -p 3306:3306 -v /data/mysql:/var/lib/mysql -e MYSQL_ROOT_PASSWORD=abcd mysql:5.7.14
```

results in `Initializing database` in the log output and instant termination without any further message. Replacing `5.7.14` by `5.7.13` results in a working MySQL instance.

When not specifying `-u 1000:1000` (and leaving `-v ...` away), it works fine.

ltangvald:
Could be a bug where something in 5.7.14 is returning a nonzero code (which will cause the entrypoint script to terminate).

ltangvald:
I tried mapping up /var/log/mysql to see if there was anything in the error log that would help, and that made it work...
The issue is probably that it can't write to the log file. Can you try also adding a -v somewhere:/var/log/mysql and see if that fixes it for you as well?

ltangvald:
We did a significant refactoring of the debian/ubuntu packaging for 5.7.14, so it's possible we no longer set the permissions for the log directory the exact same way

ffdybuster:
Yes, mapping /var/log/mysql to something the user can write to fixes the problem. Thanks a lot!

ltangvald:
Great!
Both 5.7.13 and 5.7.14 will fail without printing an error if the log directory isn't writeable, but from what I can see they both set the directory permissions the same. Will keep the issue open until we can find out why they behave differently.

ltangvald:
Found the cause: 5.7.14 changed the /etc/mysql structure to split up client and server config a bit more, and this breaks the config modifications done in Dockerfile. One of the modifications was to make the server log to stdout instead of /var/log :)

I'll make an update for it.

ffdybuster:
Cool! Thanks for fixing this :)

tianon:
Fixed in https://github.com/docker-library/mysql/pull/199. :+1:

Issue #154, opened 2016-03-26
skip-name-resolve hard coded in my.cnf


angelcervera:
I am trying to use the image in a docker compose, executing the following script from /docker-entrypoint-initdb.d : 

```
-- If not exist, create the database.
CREATE DATABASE IF NOT EXISTS sonar CHARACTER SET utf8 COLLATE utf8_general_ci;

-- Create the user and password and grant permission to access from sonar.
CREATE USER sonarUsr@localhost IDENTIFIED BY 'password';
CREATE USER sonarUsr@sonar IDENTIFIED BY 'password';
GRANT ALL ON sonar.* TO sonarUsr@localhost;
GRANT ALL ON sonar.* TO sonarUsr@sonar;
```

But the Dockerfile adds the _skip-name-resolve_ parameter in the config file. Dockerfile, line 37:

```
RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf \
    && echo 'skip-host-cache\nskip-name-resolve' | awk '{ print } $1 == "[mysqld]" && c == 0 { c = 1; system("cat") }' /etc/mysql/my.cnf > /tmp/my.cnf \
    && mv /tmp/my.cnf /etc/mysql/my.cnf
```

So, myslq can not translate the name into an IP, so it does not create the user.
Why is "_skip-name-resolve_" hard coded ?

ioggstream:
You can override that setting via command line.
Il 26/mar/2016 20:58, "Ángel Cervera Claudio" <notifications@github.com> ha
scritto:

angelcervera:
But why is this option by default?
I am concerned about the reason of this option. If it is pointedly, I am sure that there is a good reason that I unknown.
If there is not a reason, maybe the best is to keep the default mysql configuration.
If there is a reason, I am worry about to let mysql resolves names inside of the docker compose subnet in my installation.

ioggstream:
> But why is this option by default?
1. docker relies on the host DNS
2. containers uses the DNS configured in /etc/resolv.conf *at creation time*
3. if that DNS becomes unreachable (eg. disconnection, connection to a
new wifi, ...) that DNS won't work anymore
4. mysql authentication lags, waiting for client hostname resolution.

My 2¢.
Peace, R.

rocketnova:
> You can override that setting via command line.

[...]

PR #199, opened 2016-08-04
Update Dockerfile to modify correct config file

ltangvald:
MySQL 5.7.14 changed the structure of the configuration directory, so
/etc/mysql/my.cnf no longer contains the config options, but only includes other directories.
This caused the modification done to the config in Dockerfile to fail silently, creating issues
with any setups that required those changes, such as running with a different user.
Changed script to modify the new file /etc/mysql/mysql.conf.d/mysqld.cnf

Fixes issue #197

### DIFF
--- a/5.7/Dockerfile
+++ b/5.7/Dockerfile
@@ -50,9 +50,9 @@ RUN { \
 
 # comment out a few problematic configuration values
 # don't reverse lookup hostnames, they are usually another container
-RUN sed -Ei 's/^(bind-address|log)/#&/' /etc/mysql/my.cnf \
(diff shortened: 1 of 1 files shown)


yosifkit:
It looks like the default mysql install now listens publicly. We need it in the docker image, but won't that be unexpected for other users of the mysql apt repo?

Do you think we should also move the `skip-host-cache\nskip-name-resolve` to a separate config file so it is easier to remove? Hopefully to help with https://github.com/docker-library/mysql/issues/154.

ltangvald:
Yeah, that's a good point. It's more in line with the default config on other platforms, but it's potentially a breaking change and shouldn't be changed for a GA. Thanks for pointing it out :)
Moving the docker-specific options to a new file sounds like a good idea.

tianon:
Arg, `bind-address` is now back, and breaking 5.7. :lollipop:

Merging as-is to fix that! (and blaming @yosifkit appropriately for pointing it out and probably prompting you fine upstream folks to fix the issue :smile:)

PR #200, opened 2016-08-17
Handle Empty initdb.d directory

apeschel:
The docker-entrypoint.sh script does not currently handle the case where
the docker-entrypoint-initdb.d directory is empty. This commit sets the
nullglob shell option to cause the empty directory to glob to nothing.

This change will fix the following confusing message from coming up
during startup.

```
/usr/local/bin/docker-entrypoint.sh: ignoring /docker-entrypoint-initdb.d/*
```

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -eo pipefail
+shopt -s nullglob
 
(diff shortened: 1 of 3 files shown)


tianon:
Seems sane to me -- @ltangvald?

ltangvald:
LGTM. Thanks for the contribution!

PR #212, opened 2016-09-11
use "https" for communications with "mysql.com"

J0WI:
I got the following error on update.sh during testing:
`curl: (56) GnuTLS recv error (-110): The TLS connection was non-properly terminated.`
However, the script seems to work fine and curl receives all content.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -34,8 +34,8 @@ ENV MYSQL_VERSION 5.5.52
 
 # note: we're pulling the *.asc file from mysql.he.net instead of dev.mysql.com because the official mirror 404s that file for whatever reason - maybe it's at a different path?
 RUN apt-get update && apt-get install -y wget --no-install-recommends && rm -rf /var/lib/apt/lists/* \
-	&& wget "http://dev.mysql.com/get/Downloads/MySQL-$MYSQL_MAJOR/mysql-$MYSQL_VERSION-linux2.6-x86_64.tar.gz" -O mysql.tar.gz \
-	&& wget "http://mysql.he.net/Downloads/MySQL-$MYSQL_MAJOR/mysql-$MYSQL_VERSION-linux2.6-x86_64.tar.gz.asc" -O mysql.tar.gz.asc \
+	&& wget "https://cdn.mysql.com/Downloads/MySQL-$MYSQL_MAJOR/mysql-$MYSQL_VERSION-linux2.6-x86_64.tar.gz" -O mysql.tar.gz \
+	&& wget "https://cdn.mysql.com/Downloads/MySQL-$MYSQL_MAJOR/mysql-$MYSQL_VERSION-linux2.6-x86_64.tar.gz.asc" -O mysql.tar.gz.asc \
 	&& apt-get purge -y --auto-remove wget \
 	&& export GNUPGHOME="$(mktemp -d)" \
 # gpg: key 5072E1F5: public key "MySQL Release Engineering <mysql-build@oss.oracle.com>" imported
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -32,7 +32,6 @@ RUN apt-get update && apt-get install -y libaio1 pwgen && rm -rf /var/lib/apt/li
 ENV MYSQL_MAJOR 5.5
 ENV MYSQL_VERSION 5.5.52
 
-# note: we're pulling the *.asc file from mysql.he.net instead of dev.mysql.com because the official mirror 404s that file for whatever reason - maybe it's at a different path?
 RUN apt-get update && apt-get install -y ca-certificates wget --no-install-recommends && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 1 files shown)


J0WI:
Note that for MySQL 5.5 currently an insecure download is used for both, the binary and the signature.
mysql.he.net does not support https. I tired to find a host with https support for the signature, but I was not able to find one.

J0WI:
I wonder why wget does not trust dev.mysql.com. It works fine for me and I've verified on [SSL Labs](https://www.ssllabs.com/ssltest/analyze.html?d=dev.mysql.com).

For 5.6 ans 5.7 it seems that `apt-transport-https` is missing. Would it be okay to add it in this container?

tianon:
I'm personally -1 on `apt-transport-https` unless upstream recommends using `https`, since we get stronger guarantees from the GPG verification APT applies than `https` provides, but I defer to @ltangvald's preference. :+1:

J0WI:
Of course GPG is currently the best way to verify the packages, but as already mentioned we catch the signature over insecure http here. That's a very bad practice.
So an attacker could modify the package and the signature as well. It's unlikely to have a signature that matches a corrupt binary, but it's even harder when the binary is downloaded over a secure connection.
A potential risk would be if an an attacker injects an older vulnerable version of the binary and for this you also find a valid signature.

Also https is not just about security, it's also for privacy or working around network restrictions.

ltangvald:
For the apt repo, the signature is downloaded with gpg itself, so should be secure (5.6+)
https://github.com/docker-library/mysql/blob/master/5.6/Dockerfile#L30

For 5.5 we should switch to using https as suggested, though both urls in the current Dockerfile have problems with https:
https://dev.mysql.com for some reason redirects to http://cdn.mysql.com rather than https://cdn.mysql.com, while https://mysql.he.net fails.

But https://cdn.mysql.com works for both files, so I think we should have
https://cdn.mysql.com/Downloads/MySQL-$MYSQL_MAJOR/mysql-$MYSQL_VERSION-linux2.6-x86_64.tar.gz
and
https://cdn.mysql.com/Downloads/MySQL-$MYSQL_MAJOR/mysql-$MYSQL_VERSION-linux2.6-x86_64.tar.gz.asc

J0WI:
I updated this PR to only switch the url for MySQL 5.5. The host has the same problem with certs inside the container as dev.mysql.com.
Btw, is there any consistency whether to use curl or wget?

@ltangvald what is your opinion about `apt-transport-https`?

ltangvald:
@J0WI I don't think we really need it, since the download is verified with the key downloaded with gpg through apt-key.

The cert issue might have something to do with the base image contents. I'll see if I can find the cause

yosifkit:
Could the certificate error be https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812708?  Though that bug says it is fixed. :confused:

ltangvald:
@yosifkit Can you reproduce the issue? I don't get the issue, as far as I can see.
@J0WI Which version of ca-certificates have you got installed?

J0WI:
@ltangvald I have this issue only in the Travis build

tianon:
Indeed: (from https://travis-ci.org/docker-library/mysql/jobs/161303843#L715-L720)
```
converted 'https://cdn.mysql.com/Downloads/MySQL-5.5/mysql-5.5.52-linux2.6-x86_64.tar.gz' (ANSI_X3.4-1968) -> 'https://cdn.mysql.com/Downloads/MySQL-5.5/mysql-5.5.52-linux2.6-x86_64.tar.gz' (UTF-8)
--2016-09-20 11:55:41--  https://cdn.mysql.com/Downloads/MySQL-5.5/mysql-5.5.52-linux2.6-x86_64.tar.gz
Resolving cdn.mysql.com (cdn.mysql.com)... 23.207.43.221
Connecting to cdn.mysql.com (cdn.mysql.com)|23.207.43.221|:443... connected.
ERROR: The certificate of 'cdn.mysql.com' is not trusted.
ERROR: The certificate of 'cdn.mysql.com' hasn't got a known issuer.
```
:cry:

yosifkit:
I get the same error when I build the image locally.  But I run a container using `buildpack-deps:curl` and it works fine. 

Oh, I see it now.... there is no install of `ca-certificates`, just `wget`. :disappointed:

5.5/Dockerfile
```diff
@@ -34,8 +34,8 @@ ENV MYSQL_VERSION 5.5.52
 
 # note: we're pulling the *.asc file from mysql.he.net instead of dev.mysql.com because the official mirror 404s that file for whatever reason - maybe it's at a different path?
 RUN apt-get update && apt-get install -y wget --no-install-recommends && rm -rf /var/lib/apt/lists/* \
```
yosifkit: Also, that comment on line 35 is now out of date. :pray: :bow:

J0WI: What about switching to curl?

tianon: After one does `apt-get install curl`, an `apt-get purge --auto-remove curl` leaves packages around that `wget` doesn't even install, which is why we usually try to prefer `wget` for these.

ltangvald:
LGTM :)

Issue #216, opened 2016-09-26
Mysql root user is not initialized


Kern046:
Hello there !

I am facing some kind of problem with the 5.7 image. The short description is that the container is set up properly, but it is impossible to connect with the root user. My suggestion is that the root password is not initialized properly.

I am using docker-toolbox with Windows (quite sad about it but due to VPN and company group strategies it is quite hard to setup a Unix environment here) and I think Windows permissions are the problem.

I copy here my files : 

**docker-compose.yml**

```yaml
mysql:
    container_name: my_mysql_container
    build: mysql
    environment:
        - MYSQL_DATABASES=my_database
        - MYSQL_ROOT_PASSWORD=root
        - MYSQL_HOST=127.0.0.1
        - MYSQL_PORT=33306
        - MYSQL_USER=helpless
        - MYSQL_PASSWORD=dev
        - MYSQL_LOYALTY_DATABASE=my_database
    ports:
        # MySQL
        - "33306:3306"
    volumes:
        # Mysql
        - "./volumes/mysql:/var/lib/mysql"
```

**mysql/Dockerfile**
```
FROM mysql:5.7
MAINTAINER Helpless Dev <helpless.dev@wonderful_company.com>

COPY conf.d/*.cnf /etc/mysql/conf.d/
COPY docker-entrypoint-initdb.d/databases.sh /docker-entrypoint-initdb.d/databases.sh
COPY docker-entrypoint-initdb.d/config.sh /docker-entrypoint-initdb.d/config.sh

RUN chmod a+x /docker-entrypoint-initdb.d/config.sh && chown root:root /docker-entrypoint-initdb.d/config.sh
RUN chmod a+x /docker-entrypoint-initdb.d/databases.sh && chown root:root /docker-entrypoint-initdb.d/databases.sh
```

**mysql/conf.d/custom.cnf**

```
[mysqld]
default-storage-engine=INNODB
init-connect='SET NAMES utf8'
character-set-server=utf8
collation-server=utf8_unicode_ci
bind-address		= 0.0.0.0
skip-external-locking
key_buffer		= 16M
max_allowed_packet	= 16M
thread_stack		= 192K
thread_cache_size       = 8
query_cache_limit	= 1M
query_cache_size        = 16M
expire_logs_days	= 10
max_binlog_size         = 100M

[mysqldump]
quick
quote-names
max_allowed_packet	= 16M

[isamchk]
key_buffer		= 16M
```

[...]

PR #218, opened 2016-09-30
Catch errors when scraping datadir from help output

yosifkit:
Before this change, a user would just get the container exiting and zero output.  Now a user will get some friendly output about the command that was run and then the error messages that `mysqld` emitted.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -18,12 +18,27 @@ for arg; do
 	esac
 done
 
+_check_config() {
+	toRun=( "$@" --verbose --help --log-bin-index="$(mktemp -u)" )
+	if ! errors="$("${toRun[@]}" 2>&1 >/dev/null)"; then
+		cat >&2 <<-EOM
+
+			ERROR: mysqld failed while attempting to determine datadir
+			command was: "${toRun[*]}"
+
+			$errors
+		EOM
+		exit 1
+	fi
+}
+
 _datadir() {
-	"$@" --verbose --help --log-bin-index=`mktemp -u` 2>/dev/null | awk '$1 == "datadir" { print $2; exit }'
+	"$@" --verbose --help --log-bin-index="$(mktemp -u)" 2>/dev/null | awk '$1 == "datadir" { print $2; exit }'
 }
 
 # allow the container to be started with `--user`
 if [ "$1" = 'mysqld' -a -z "$wantHelp" -a "$(id -u)" = '0' ]; then
+	_check_config "$@"
 	DATADIR="$(_datadir "$@")"
 	mkdir -p "$DATADIR"
 	chown -R mysql:mysql "$DATADIR"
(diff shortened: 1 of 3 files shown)


yosifkit:
Related to https://github.com/docker-library/mysql/issues/216#issuecomment-250813336

tianon:
LGTM!

cc @ltangvald

ltangvald:
Maybe separate it from the datadir fetch? We have a commit for this in the upstream image, actually:
https://github.com/mysql/mysql-docker/commit/05833afcfb62f88fc294d1afed584cec1d5bdd0e

yosifkit:
I'm not sure I understand the need to separate the error detection from the datadir scraping.  This code ensures we only have to run `mysqld --verbose --help` once to get the error and datadir.  Plus, now that we have the `_datadir` function this ensures that the command is checked for errors any time it is called (which is currently two times, once as root, [line 27/37](https://github.com/docker-library/mysql/pull/218/files#diff-b571e00bf9d60d273fa70dfc49255a2eR37), and once as mysql user, [line 35/45](https://github.com/docker-library/mysql/pull/218/files#diff-b571e00bf9d60d273fa70dfc49255a2eR45))

ltangvald:
It's somewhat a personal preference since I think it's easier to read and understand the code, but there is some benefit to keeping different logical operations (fetch datadir and check for errors) separate, since it makes future debugging and maintenance simpler.

yosifkit:
That makes sense, though the only problem I have with separating them is that they could become out of sync.  The reason for checking for errors is to make sure the datadir scraping will succeed.

As an alternative solution, we could just ignore the failure and add an `|| true`.  Aside from the non-zero exit, are there any `mysqld --verbose -help` errors that would cause the datadir scrape to fail?  Perhaps we just allow it to fail later when it does `mysql_install_db` or `--initialize-insecure`.

ltangvald:
Yeah, we added it as a more general error checking to fail fast and notify users if e.g. a custom config contains invalid entries, not specifically for the datadir scrape. I think it's a useful thing to have, since it lets us print a decent error message in such cases.

Can't really think of anything else that would cause the scrape to fail.

yosifkit:
Updated to separate config error check and datadir scraping.

5.5/docker-entrypoint.sh
```diff
@@ -18,19 +18,36 @@ for arg; do
 	esac
 done
 
+_check_config() {
+	toRun=( "$@" --verbose --help --log-bin-index="$(mktemp -u)" )
+	if ! errors="$("${toRun[@]}" 2>&1 >/dev/null)"; then
+		cat >&2 <<-EOM
+
+			ERROR: mysqld failed while attempting to determine datadir
```
yosifkit: I need to update this text to something more appropriate.

ltangvald:
LGTM

tianon:
:+1:

PR #222, opened 2016-10-17
Specify host for client login during init to ignore MYSQL_HOST value

ltangvald:
When the server is started up during initialization, it will only allow access
through the localhost socket interface. The entrypoint script logs in and changes
the root user to allow remote access, but if the MYSQL_HOST environment variable is
defined it will be used by the client, causing the script's login to fail if it
is set to anything other than localhost.

Adding a -hlocalhost to the client login during init makes it ignore the environment variable.

This should prevent problems as seen in issue #216


ltangvald:
Note that MYSQL_HOST overriding --socket=protocol is a bug in the server: https://bugs.mysql.com/bug.php?id=31832

PR #232, opened 2016-10-31
Move docker-specific [mysqld] additions to their own "docker.cnf" file

tianon:
This was discussed a bit over in https://github.com/docker-library/mysql/pull/199#issuecomment-237620674.  The main benefit is that it makes it much easier to remove the Docker additions entirely if so desired. :+1:


ltangvald:
LGTM :)

PR #236, opened 2016-11-15
Added environment variable to run SQL command at initialization phase

robvadai:


yosifkit:
What functionality is gained that is not handled by just adding a `custom.sql` to `/docker-entrypoint-initdb.d/`?  I'd rather not complicate the image with another environment variable without a strong use case.

Thoughts? @ltangvald or @tianon

robvadai:
@yosifkit if you define a bash script you have to construct the full mysql command with username etc. Using this environment variable saves you doing all this.

yosifkit:
You don't need a bash script; it'll import sql and sql.gz files directly: https://github.com/docker-library/mysql/blob/958442857411934f3f3a525325a2e10c7e8e8155/8.0/docker-entrypoint.sh#L128-L129.

ltangvald:
Yeah, I don't think this adds anything that isn't already supported (plus it's only been added for 5.5)

robvadai:
I got it. It should be added to the README then to make it clear. Thanks guys

PR #237, opened 2016-11-21
Add "file_env" support, especially for Docker secrets

tianon:
This adds explicit support for the following:

- `MYSQL_ROOT_PASSWORD_FILE`
- `MYSQL_DATABASE_FILE`
- `MYSQL_USER_FILE`
- `MYSQL_PASSWORD_FILE`

See also:

- docker-library/wordpress#186
- docker-library/postgres#225
- docker-library/redmine#43

I've only updated `8.0/docker-entrypoint.sh` here so that this can serve as a straw-man for discussion -- once the discussion concludes, I'll update the PR with the result and push the functionality across the board to all supported versions. :+1:

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -18,6 +18,28 @@ for arg; do
 	esac
 done
 
+# usage: file_env VAR [DEFAULT]
+#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
+# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
+#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
+file_env() {
+	local var="$1"
+	local fileVar="${var}_FILE"
+	local def="${2:-}"
+	if [ "${!var:-}" ] && [ "${!fileVar:-}" ]; then
+		echo >&2 "error: both $var and $fileVar are set (but are exclusive)"
+		exit 1
+	fi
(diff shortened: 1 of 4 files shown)


tianon:
I saw this was added to `mysql/mysql-server` in https://github.com/mysql/mysql-docker/commit/0473555536990a99a47b880253915d6b45489567, but IMO it's cleaner to add an explicit "get the password from this specific file" option than to overload the existing option.  For example, if someone uses a bad `-w`/`--workdir` on their container, their password might be accidentally interpreted as a file to read, and there will be no external indication that it's happened and the admin will be scratching their head wondering why their password isn't working. :sweat_smile: 

(That being said, we defer to you, @ltangvald -- however you think this should be implemented here is the direction we'll go. :+1:)

tianon:
cc @cyli @mstanleyjones FYI

ltangvald:
Hm, yeah. I'm not 100% sure why we used the same variable (other than a general desire to keep the number of extra env variables down). We considered the chance of accidentally setting the password to a valid file path low, but I agree it's cleaner to make it explicit.
The file_env function looks a bit complex to me, but it's nice to have the functionality and I can't think of a simpler way off the top of my head, so overall I think this looks good :)

tianon:
Yeah, the goal of the function was to abstract the behavior in a clean, generic way so we don't have "this or that" logic peppered all over the file (which is error prone on top of verbose). :sweat_smile:

I've updated to push this change to all the versions now. :smile: :+1:

PR #249, opened 2016-12-31
Add ability to change host value for root user to secure container

### LINKED ISSUE #231: Increase security by adding possibility to have 'root'@'localhost' and not only 'root'@'%'
Hi,

I would like to increase the security in my mySQL docker installation by not allowing root to login from any computer that can access the port on the host.

I am used to that root only is allowed to login from localhost. Since we are talking about a docker container here I guess there can be arguments that one does not in all installations have access to the container itself and need to act as root from another computer.

I would however prefer if there was an option not to have root@% as the default and not the only option when installing.

If you have access to the computer hosting the docker container one can use the command "docker exec -it mysql mysql -u root -p" to run a mySQL client and access the server from localhost.

I was looking into adding files to the /docker-entrypoint-initdb.d/ folder but these files are executed before the commands related to the option MYSQL_ONETIME_PASSWORD is executed. Since the later rely on 'root'@'%' it is not possible to change from @'%' to @'localhost' in a script in this folder.

I see a number of ways to fix this.

1) Add an option MYSQL_ROOT_HOST that have a default value of '%' (for backwards compatibility) but gives the user of the container the option to set it to 'localhost' or possibly another name that is used by the computer connecting to the docker container.

2) Move the handling of the MYSQL_ONETIME_PASSWORD above the handling of docker-entrypoint-initdb.d. A user can then add .sql commands to it that changes the '%' for root to 'localhost' or whatever he/she prefers.

I am sure there are more options.

My proposal would be #1 above since I think this will highlight the possible security risk with having 'root'@'%' and give the user of the container the option to increase the security.

If we agree on a way forward I can create a pull-request.

/Morgan

yosifkit:
Now you can easily remove remote MySQL root access or secure to a specific IP address (DNS name would require custom cnf to override the provided config files).  `root@localhost` will always work regardless of this new value.

```console
$ docker run -d --name mysql -e MYSQL_ROOT_PASSWORD=pass -e MYSQL_ROOT_HOST=localhost -e MYSQL_DATABASE=db -e MYSQL_USER=user -e MYSQL_PASSWORD=pass2 mysql:5.6
$ docker run -it --rm --link mysql mysql:5.7 mysql -hmysql -uroot -ppass
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'172.17.0.5' (using password: YES)
$ docker run -it --rm --link mysql mysql:5.7 mysql -hmysql -uuser -ppass2 db
mysql: [Warning] Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 12
Server version: 5.6.35 MySQL Community Server (GPL)

Copyright (c) 2000, 2016, Oracle and/or its affiliates. All rights reserved.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> ^DBye
```

fixes #231

Borrows heavily from https://github.com/mysql/mysql-docker/commit/6fbb227bae9ba468404a96f3aa112012d9a8bb6d

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -113,14 +113,28 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			export MYSQL_ROOT_PASSWORD="$(pwgen -1 32)"
 			echo "GENERATED ROOT PASSWORD: $MYSQL_ROOT_PASSWORD"
 		fi
+
+		rootCreate=
+		# default root to listen for connections from anywhere
+		file_env 'MYSQL_ROOT_HOST' '%'
+		if [ ! -z "$MYSQL_ROOT_HOST" -a "$MYSQL_ROOT_HOST" != 'localhost' ]; then
+			# no, we don't care if read finds a terminating character in this heredoc
+			# https://unix.stackexchange.com/questions/265149/why-is-set-o-errexit-breaking-this-read-heredoc-expression/265151#265151
+			read -r -d '' rootCreate <<-EOSQL || true
+				CREATE USER 'root'@'${MYSQL_ROOT_HOST}' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
+				GRANT ALL ON *.* TO 'root'@'${MYSQL_ROOT_HOST}' WITH GRANT OPTION ;
+			EOSQL
+		fi
+
 		"${mysql[@]}" <<-EOSQL
 			-- What's done in this file shouldn't be replicated
 			--  or products like mysql-fabric won't work
 			SET @@SESSION.SQL_LOG_BIN=0;
 
-			DELETE FROM mysql.user ;
-			CREATE USER 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
-			GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION ;
+			DELETE FROM mysql.user WHERE user NOT IN ('mysql.sys', 'mysqlxsys', 'root') OR host NOT IN ('localhost') ;
+			SET PASSWORD FOR 'root'@'localhost'=PASSWORD('${MYSQL_ROOT_PASSWORD}') ;
+			GRANT ALL ON *.* TO 'root'@'localhost' WITH GRANT OPTION ;
+			${rootCreate}
 			DROP DATABASE IF EXISTS test ;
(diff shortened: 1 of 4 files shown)


5.5/Dockerfile
```diff
@@ -68,6 +68,9 @@ RUN mkdir -p /var/lib/mysql /var/run/mysqld \
 # ensure that /var/run/mysqld (used for socket and lock files) is writable regardless of the UID our mysqld instance ends up having at runtime
 	&& chmod 777 /var/run/mysqld
 
+# default root to listen for connections from anywhere
+ENV MYSQL_ROOT_HOST %
```
tianon: This is going to cause issues with `file_env` (`-e MYSQL_ROOT_HOST_FILE=...` will lead to `error: both MYSQL_ROOT_HOST and MYSQL_ROOT_HOST_FILE are set (but are exclusive)`), so needs to be specified as the default in the script instead (via the second argument to `file_env`).

yosifkit:
Updated to work properly with `MYSQL_ROOT_HOST_FILE`; cannot use an empty value to remove the `root@%` user, you must use `localhost` or `MYSQL_ROOT_HOST_FILE=/dev/null` or another empty file.

5.5/docker-entrypoint.sh
```diff
@@ -113,14 +113,28 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			export MYSQL_ROOT_PASSWORD="$(pwgen -1 32)"
 			echo "GENERATED ROOT PASSWORD: $MYSQL_ROOT_PASSWORD"
 		fi
+
+		rootCreate=
+		# default root to listen for connections from anywhere
+		file_env 'MYSQL_ROOT_HOST' '%'
+		if [ ! -z "$MYSQL_ROOT_HOST" -a "$MYSQL_ROOT_HOST" != 'localhost' ]; then
```
yosifkit: Should this check also look for loopback IP addresses like `127.0.0.1` or `::1`?

yosifkit: Found the answer: No; `localhost` for mysql is just the unix socket, so IP addresses do not factor in to it.

yosifkit:
cc @ltangvald

ltangvald:
Looks good.

PR #259, opened 2017-01-26
Add "openssl" and invoke "mysql_ssl_rsa_setup" at runtime so that automatic SSL works the same as the MySQL packages

### LINKED ISSUE #257: SSL Certificates generation
MySQL, starting from 5.7.5 version, automatically generates ssl keys and certificates and stores them in data directory. 

I have installed mysql 5.7.17 using mysql docker image and I do not see certificates being generated in the default data directory. 

Thanks in advance.

tianon:
Fixes #257

cc @ltangvald

I verified this by building/running with no extra flags, then connecting from a separate container using `--ssl-mode REQUIRED` (to force SSL communications). :+1:

### DIFF
--- a/5.7/Dockerfile
+++ b/5.7/Dockerfile
@@ -19,12 +19,18 @@ RUN set -x \
 
 RUN mkdir /docker-entrypoint-initdb.d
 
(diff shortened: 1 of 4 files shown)


ltangvald:
Great!

Issue #247, opened 2016-12-29
mysql:5.5 Can't connect to local MySQL server through socket '/tmp/mysql.sock'


hxzqlh:
I once pull `mysql:5.5`(5.5.53) to my docker image repository, before `mysql:5.5` updating to `5.5.54`, the `Dockerfile` keep unchanged like this:

```
FROM mysql:5.5
ENV MYSQL_ROOT_PASSWORD 123
ENV MYSQL_USER vdna
ENV MYSQL_PASSWORD 123
USER root
COPY init_db.sh /docker-entrypoint-initdb.d/init_db.sh
``` 

`init_db.sh`:
```
#!/bin/sh

mysql -uroot -p$MYSQL_ROOT_PASSWORD -e "grant all privileges on *.* to '$MYSQL_USER'@'%' identified by '$MYSQL_PASSWORD';"
mysql -uroot -p$MYSQL_ROOT_PASSWORD -e "flush privileges;"
```

build the Dockerfile to image:

```
docker build -t hxzqlh/mysql_db .
```

when run the container:`docker run -it --rm hxzqlh/mysql_db`,  `5.5.54` version cannot start container , from the console output, problem occured during executing the `init_db.sh`.

```
Initializing database
161229  2:41:05 [Note] Ignoring --secure-file-priv value as server is running with --bootstrap.
161229  2:41:05 [Note] /usr/local/mysql/bin/mysqld (mysqld 5.5.54) starting as process 67 ...
161229  2:41:05 [Note] Ignoring --secure-file-priv value as server is running with --bootstrap.
161229  2:41:05 [Note] /usr/local/mysql/bin/mysqld (mysqld 5.5.54) starting as process 73 ...

PLEASE REMEMBER TO SET A PASSWORD FOR THE MySQL root USER !
To do so, start the server, then issue the following commands:

/usr/local/mysql/bin/mysqladmin -u root password 'new-password'
/usr/local/mysql/bin/mysqladmin -u root -h  password 'new-password'

Alternatively you can run:
/usr/local/mysql/bin/mysql_secure_installation

which will also give you the option of removing the test
databases and anonymous user created by default.  This is
strongly recommended for production servers.

See the manual for more instructions.

Please report any problems at http://bugs.mysql.com/

Database initialized
161229  2:41:05 [Note] --secure-file-priv is set to NULL. Operations related to importing and exporting data are disabled
161229  2:41:05 [Note] mysqld (mysqld 5.5.54) starting as process 77 ...
[...]

PR #266, opened 2017-02-16
Fetch client --socket value from config during init

ltangvald:
Fetch client --socket value from config instead of hardcoding it
    
Hardcoding the socket path for the init run broke initdb.d/ scripts that
expected a different location (because 5.5 has a different default or the user
changed the config).
Added a function for fetching it with mysqld --verbose --help as we do for
datadir, and use that for the initialization runs of the server and client.
Should fix issue #247

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -54,19 +54,18 @@ _check_config() {
 	fi
 }
 
-_datadir() {
-	"$@" --verbose --help --log-bin-index="$(mktemp -u)" 2>/dev/null | awk '$1 == "datadir" { print $2; exit }'
-}
-
-# For use with the client if user passes the --socket argument
-_socket() {
-	"$@" --verbose --help --log-bin-index="$(mktemp -u)" 2>/dev/null | awk '$1 == "socket" { print $2; exit }'
+# Fetch value from server config
+# We use mysqld --verbose --help instead of my_print_defaults because the
+# latter only show values present in config files, and not server defaults
+_get_config() {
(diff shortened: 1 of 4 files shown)


5.5/docker-entrypoint.sh
```diff
@@ -58,6 +58,11 @@ _datadir() {
 	"$@" --verbose --help --log-bin-index="$(mktemp -u)" 2>/dev/null | awk '$1 == "datadir" { print $2; exit }'
 }
 
+# For use with the client if user passes the --socket argument
+_socket() {
+	"$@" --verbose --help --log-bin-index="$(mktemp -u)" 2>/dev/null | awk '$1 == "socket" { print $2; exit }'
+}
```
tianon: Do you think there'd be any value in making something like a `_get_config` function to normalize these two?  I guess `_check_config` is pretty similar too.

Something like this is what I was kind of thinking:
```bash
_get_config() {
	local conf="$1"; shift
	"$@" --verbose --help --log-bin-index="$(mktemp -u)" 2>/dev/null | awk '$1 == "'"$conf"'" { print $2; exit }'
}

datadir="$(_get_config 'datadir' "$@")"
socket="$(_get_config 'socket' "$@")"
```

(will defer to your decision, just wanted to throw the idea out there in case you hadn't considered it :+1:)

ltangvald: Looks to be working fine. I'll get the pr update up tomorrow once it's applied and tested for all versions.

tianon:
Cool :+1:

PR #282, opened 2017-04-11
Use new -core packages for 8.0.1+

ltangvald:
Should reduce the image size by about 200MB, but may need to discuss some points below:
* secure-file-priv:
  https://dev.mysql.com/doc/refman/5.7/en/server-options.html#option_mysqld_secure-file-priv
  The setting restricts where the server can read and write on the filesystem with LOAD DATA and SELECT INTO OUTFILE. The default is /var/lib/mysql-files, but I set it to NULL, which disables these functions
* Binaries:
  The -core packages contain the same binaries we have in the upstream docker images, and not all of the binaries in the standard server images are included (since 8.0 is still dmr we can change this in the future if we realize we're missing something that people need)
* Config:
  Since the core packages have no config we need to generate them from scratch. I added them to the git tree instead of generating in the script, since I think it's a bit easier for people to see exactly what the config is that way. Maybe change directory structure from mysql/ to config/ or config/mysql/?
Commit message:
-------------------
The packages mysql-community-client-core and mysql-community-server-core
contain only the binaries needed by the client and server, with no
configuration, services or debug binaries, making them significantly
smaller in size.

* Install -core packages instead of full mysql-server
  Images that build on this and install further packages that depend on
  MySQL may lose the size benefit if the dep is on "mysql-server", but
  should otherwise work as before.
* Add all config files, since core packages do not include any.
  Included the config files in git instead of generating them in the build
  script, to make it a bit easier to see exactly what the default config
  of the image is.


tianon:
Very, very interesting! :smile: :+1:

Do you think that it would be worthwhile to simplify the configuration structure while we're at it, given that we have to provide it all?  Or is it more valuable to stick to the same structure/content as the "full" releases do (as you've done here)?

ltangvald:
Not entirely sure. Keeping it the same makes it a bit easier for people to transition straight from full releases to Docker, the structure has some issues (like mysql.conf.d/ being loaded after conf.d/, so you can't put anything in conf.d/ to override any of the settings in mysql.conf.d/mysqld.cnf).

I think I like the idea of simplifying. The structure we use in the full packages was largely created to be compatible with native packages (which are like that to be "compatible" with MariaDB and other variants). I'll update the pr :)

ltangvald:
I've changed it so my.cnf now contains the default server config, and only includes conf.d. Should actually make our docs more correct :)
I've left docker.cnf as it is, since the flags it sets don't take values, so you have to actually remove/change the file to override them.

tianon:
:metal: (going to look/reply on https://github.com/docker-library/mysql/issues/289 now :innocent:)

PR #288, opened 2017-04-25
Fix behavior during initialization timeout

fpytloun:
Add configurable timeout for MySQL start during initialization. On timeout, remove datadir to allow new initialization try on next container creation. Otherwise mysql will end up in empty unconfigured state without root password set, etc.

Better fix for: https://github.com/docker-library/mysql/pull/159

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -97,7 +97,7 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 
 		mysql=( mysql --protocol=socket -uroot -hlocalhost --socket="${SOCKET}" )
 
-		for i in {30..0}; do
+		for i in $(seq ${MYSQL_START_TIMEOUT:-30} -1 0); do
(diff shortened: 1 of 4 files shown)


NiR-:
@tianon @yosifkit sorry to ping you directly, but is there any chance this PR got merged sooner or later, or is there any specific issue that should be addressed first?

I'm actually facing failing MySQL init process due to "big" `innodb_log_file_size` value and I really would like to increase timeout delay rather than decrease that config value.

yosifkit:
Sorry for the long tail. :bowing_man:

Making the timeout configurable could be useful. Though should probably be addressed after or in #471.

Even though it is currently gated by only running on an empty directory (well, partly-initialized), I would rather not have the `rm -rf $DATADIR/mysql`. With the possible changes in #471, the temporary localhost-only server could be started without it being a new database; which means a slow start would delete the data.

yosifkit:
Apologies for the delay :bowing_man:

Closing in favor of https://github.com/docker-library/mysql/issues/646

PR #294, opened 2017-05-04
Update to 5.5.56 (accounting for the tarball rename)

tianon:
cc @ltangvald

(Build was failing, so I checked https://dev.mysql.com/downloads/mysql/5.5.html#downloads and sure enough, tarballs have been renamed for this release :sweat_smile:)


ltangvald:
Whoops. My bad, sorry. 5.5.56 only contains a change for a build flag, so wouldn't affect published binaries, but I forgot to consider that 5.5.55 would of course be removed from the repo :)

PR #347, opened 2017-12-03
Prevent mysql.session account deletion instead of mysqlxsys

DmitryErshov:
Changes in MySQL 5.7.19
https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-19.html

> #### Account Management Notes
> During data directory initialization or upgrade, MySQL now creates a **'mysql.session'@'localhost'** reserved account. This account is used internally by plugins to access the server.

> Plugins such as Group Replication and X Plugin now use the **mysql.session** account added in this version.

### DIFF
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -143,7 +143,7 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			--  or products like mysql-fabric won't work
 			SET @@SESSION.SQL_LOG_BIN=0;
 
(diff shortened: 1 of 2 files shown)


ltangvald:
This is how the upstream image currently works (https://github.com/mysql/mysql-docker/blob/mysql-server/5.7/docker-entrypoint.sh#L119)
However, it might be best to simply drop user deletion for 5.7+, since 5.7 and 8.0 do not create the "junk" users from 5.6 and older.

The 5.7 image will by default only have root@localhost, root@% (because entrypoint creates it), mysql.session and mysql.sys.

ltangvald:
Merged over as-is for now, and will look at changing/removing the deletion separately.

PR #348, opened 2017-12-07
Update docker-entrypoint to add database_wildcard

GuigZ-:
Hi,

In my project, i  need a database_wildcard to create some databases.

I create this PR if you want to add wildcard in your container.

BR GuigZ-


tianon:
I'm not entirely sold on the additional complexity added here relative to how often it's necessary -- IMO, this would be better suited by users dropping a one-liner into `/docker-entrypoint-initdb.d` for the use cases that require this.

@ltangvald your thoughts?

GuigZ-:
Hi,
Also if you prefer to check if wildcard is in the database name is possible too.

What do you think?

neumayer:
I also think that the /docker-entrypoint-initdb.d suggestion is more generic and as such preferable.

tianon:
I'm still strongly of the opinion that this sort of functionality ought to be served via user-supplied files in `/docker-entrypoint-initdb.d` instead.

GuigZ-:
Okay, i use a file in `/docker-entrypoint-initdb.d`, thanks.

PR #358, opened 2017-12-21
Pass extra mysqld flags to mysql_install_db

### LINKED ISSUE #356: Setting ``--innodb-log-file-size=1G`` leads to initialization failure, unable to connect as root
I'm switching from a configuration file to a command line configuration for MySQL 5.6, using this ``docker-compose.yml`` section:

```
version: '2.1'
services:
  mysql:
    image: mysql:5.6
    environment:
      - MYSQL_ROOT_PASSWORD=my_password
    command:
      - --character-set-server=utf8
      - --collation-server=utf8_general_ci
      - --innodb-buffer-pool-size=1G
      - --innodb-flush-log-at-trx-commit=0
      - --innodb-log-file-size=1G
      - --max-allowed-packet=100M
    volumes:
      - mysqlvolume:/var/lib/mysql
volumes:
    mysqlvolume:
```
The ``docker-entrypoint.sh`` initialization script will create the database (``mysql_install_db --datadir="$DATADIR" --rpm --keep-my-cnf``) with the default InnoDB buffer pool (48 MB). It will then start ``mysqld`` and wait 30 seconds for it to initialize.  This initialization includes recreating the buffer pool, which takes longer than 30 seconds in TravisCI ([failed build](https://travis-ci.org/mozilla/kuma/jobs/319777072)).  ``docker-entrypoint.sh`` exits early, and the root password remains unset. This leads to the follow-on errors, such as  ``ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)`` and ``Host '172.18.0.9' is not allowed to connect to this MySQL server``.

This does not happen when a configuration file is used, because ``mysql_install_db`` will initialize with the desired InnoDB buffer pool size.

Some options are:

- Pass options to ``mysql_install_db`` as well
- Increase the timeout from 30 seconds to a larger value

tianon:
Fixes #356

I've tested and verified this with `--datadir=/data/db --innodb-log-file-size=1G` specified at runtime, and everything works (on both `mysql:5.5` and `mysql:5.6`), with the appropriate and expected output from `mysql_install_db` while initializing InnoDB:

```
...
171221 21:35:35  InnoDB: Setting file ./ibdata1 size to 10 MB
InnoDB: Database physically writes the file full: wait...
171221 21:35:35  InnoDB: Log file ./ib_logfile0 did not exist: new to be created
InnoDB: Setting log file ./ib_logfile0 size to 1024 MB
InnoDB: Database physically writes the file full: wait...
InnoDB: Progress in MB: 100 200 300 400 500 600 700 800 900 1000
MySQL init process in progress...
MySQL init process in progress...
171221 21:35:37  InnoDB: Log file ./ib_logfile1 did not exist: new to be created
InnoDB: Setting log file ./ib_logfile1 size to 1024 MB
InnoDB: Database physically writes the file full: wait...
InnoDB: Progress in MB: 100 200 300 400 500 600 700 800 900 1000
MySQL init process in progress...
...
```

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -88,7 +88,8 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 		mkdir -p "$DATADIR"
 
 		echo 'Initializing database'
-		mysql_install_db --datadir="$DATADIR" --rpm --basedir=/usr/local/mysql
+		# "Other options are passed to mysqld." (so we pass all "mysqld" arguments directly here)
+		mysql_install_db --datadir="$DATADIR" --rpm --basedir=/usr/local/mysql "${@:2}"
 		echo 'Database initialized'
 
 		SOCKET="$(_get_config 'socket' "$@")"
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -88,7 +88,8 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 		mkdir -p "$DATADIR"
 
 		echo 'Initializing database'
-		mysql_install_db --datadir="$DATADIR" --rpm --keep-my-cnf
+		# "Other options are passed to mysqld." (so we pass all "mysqld" arguments directly here)
+		mysql_install_db --datadir="$DATADIR" --rpm --keep-my-cnf "${@:2}"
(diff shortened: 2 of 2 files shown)


tianon:
(Used a custom `--datadir` in my testing to verify that this doesn't break any of that functionality, and I also verified that the containers were in fact writing to `/data/db` as a result.)

PR #376, opened 2018-02-15
Update 8.0 to Debian Stretch

### LINKED ISSUE #367: Offer image variants that are based on debian:*-slim
Hi!

Currently the mysql images use `debian:jessie` as base, which is 50MB. It would be great to add `mysql:*-slim` image variants based on `debian:jessie-slim` (which is only 30MB) or even `debian:jessie-stretch` (which is only 22MB) to reduce the overall image size.

Would a PR for this be accepted? :-)

### LINKED ISSUE #375: Update MySQL 8 image to 8.0.4
MySQL 8.0.4 has been released on January 23: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-4.html

The MySQL 8 image should be updated.

tianon:
Closes #375
Closes #367

This should be reasonably safe to do given that 8.0 is still in pre-release stages.

It looks like we should also consider updating both 5.6 and 5.7 to Stretch, but I didn't want to rock that boat in the same breath as getting 8.0.4 out. :smile:

### DIFF
--- a/8.0/Dockerfile
+++ b/8.0/Dockerfile
@@ -1,6 +1,9 @@
-FROM debian:jessie
+FROM debian:stretch-slim
+
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN groupadd -r mysql && useradd -r -g mysql mysql
+
+RUN apt-get update && apt-get install -y --no-install-recommends gnupg dirmngr && rm -rf /var/lib/apt/lists/*
 
 # add gosu for easy step-down from root
 ENV GOSU_VERSION 1.7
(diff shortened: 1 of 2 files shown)


ltangvald:
LGTM. I think stretch-slim is stripped in a similar manner to the ol7-slim image we use upstream.
+1 on updating the 5.6 and 5.7 as well, but as you mention there's no need to do it right now.

edmorley:
Many thanks for doing this - it reduced the `mysql:8` image by 52MB :-)
(Before: [343MB](https://travis-ci.org/docker-library/mysql/jobs/329160359#L1021), after: [291MB](https://travis-ci.org/docker-library/mysql/jobs/341991239#L1115))

Shall I open a new ticket to track doing the same to 5.7 et al, or would reopening #367 make more sense?

tianon:
https://github.com/docker-library/mysql/pull/382 :+1:

Issue #401, opened 2018-04-09
MySQL:8.0 is taking unusual time to Initialize


the-redback:
`mysql:8.0` (aka `8.0.4`) is taking unusual time (approximately 10+ minutes) to initialize in docker (local machine). 

If you look closely at log's time, the last line took 7 minutes to execute.

```console
Initializing database
2018-04-09T10:00:29.926078Z 0 [Warning] [MY-011070] Disabling symbolic links using --skip-symbolic-links (or equivalent) is the default. Consider not using this option as it is deprecated and will be removed in a future release.
2018-04-09T10:00:29.926165Z 0 [System] [MY-010116] /usr/sbin/mysqld (mysqld 8.0.4-rc) starting as process 30 ...
mbind: Operation not permitted
mbind: Operation not permitted
mbind: Operation not permitted
mbind: Operation not permitted
2018-04-09T10:00:46.842725Z 0 [Warning] [MY-010068] CA certificate ca.pem is self signed.
2018-04-09T10:00:46.903128Z 5 [Warning] [MY-010453] root@localhost is created with an empty password ! Please consider switching off the --initialize-insecure option.
2018-04-09T10:07:28.643932Z 5 [Warning] [MY-010315] 'user' entry 'mysql.infoschema@localhost' ignored in --skip-name-resolve mode.
``` 

For your Information, in this same machine, `mysql:8.0.3` doesn't take much time at all to Initialize the Database. 

Here is my configuration:

**OS:** ubuntu 16.04

```console
$ docker version
Client:
 Version:	17.12.1-ce
 API version:	1.35
 Go version:	go1.9.4
 Git commit:	7390fc6
 Built:	Tue Feb 27 22:17:40 2018
 OS/Arch:	linux/amd64

Server:
 Engine:
  Version:	17.12.1-ce
  API version:	1.35 (minimum version 1.12)
  Go version:	go1.9.4
  Git commit:	7390fc6
  Built:	Tue Feb 27 22:16:13 2018
  OS/Arch:	linux/amd64
  Experimental:	false
```

Process to recreate and full log:

```console
$ docker pull mysql:8.0
8.0: Pulling from library/mysql
2a72cbf407d6: Pull complete 
38680a9b47a8: Pull complete 
4c732aa0eb1b: Pull complete 
c5317a34eddd: Pull complete 
f92be680366c: Pull complete 
e8ecd8bec5ab: Pull complete 
2a650284a6a8: Pull complete 
[...]

PR #404, opened 2018-04-19
Run update.sh

TimWolla:
Apparently docker-library-bot is broken for this repository.


TimWolla:
Okay, the Travis failure probably explains why docker-library-bot is broken.

The issue seems to be reported in #401.

yosifkit:
@tianon saw the bot failures and made the fix in https://github.com/docker-library/mysql/pull/406

PR #406, opened 2018-04-19
Fix a few minor issues with docker-entrypoint.sh for 8.0.11

tianon:


tianon:
Here's an example of output from `mysqld --help --verbose` that tripped up our `awk`:

```
  --mysqlx-port-open-timeout[=#] 
                      How long X Plugin is going to retry binding of server
                      socket (in case of failure)
```

8.0/docker-entrypoint.sh
```diff
@@ -143,7 +145,7 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			--  or products like mysql-fabric won't work
 			SET @@SESSION.SQL_LOG_BIN=0;
 
-			SET PASSWORD FOR 'root'@'localhost'=PASSWORD('${MYSQL_ROOT_PASSWORD}') ;
+			ALTER USER 'root'@'localhost' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
```
tianon: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-11.html#mysqld-8-0-11-deprecation-removal

> The `PASSWORD()` function. Additionally, `PASSWORD()` removal means that `SET PASSWORD ... = PASSWORD('auth_string')` syntax is no longer available.

tianon: (Using `ALTER USER` instead is what's done in `mysql/mysql-server`: https://github.com/mysql/mysql-docker/blob/6c4a6defc2f74a0b4b5eb54847f849463b92d5bb/8.0/docker-entrypoint.sh#L111)

tianon:
8.0 is green, the rest of the Travis builds are unrelated (transitory) failures :+1:

tianon:
Probably also worth noting here the following commits:

- ff62794 Move "latest" to 8.0 now that it's GA
- 6662ac7 Update to 5.7.22-1debian9
- d32482c Update to 5.5.60
- e279ec9 Update to 5.6.40-1debian9

PR #416, opened 2018-04-25
Automatically run mysql_upgrade for 8.0 image

ltangvald:
It is recommended to always run mysql_upgrade when upgrading
MySQL, and it is required when upgrading from an older major version.

Entrypoint script checks if mysql_upgrade has previously been run on the
current version. If not it does a temporary startup and runs it.

---

This should fix most issues related from latest tag having changed from 5.7 to 8.0

### DIFF
--- a/8.0/docker-entrypoint.sh
+++ b/8.0/docker-entrypoint.sh
@@ -64,6 +64,36 @@ _get_config() {
 	# match "datadir      /some/path with/spaces in/it here" but not "--xyz=abc\n     datadir (xyz)"
 }
 
+_stop_server(){
+	# This isn't ideal, but we can generally assume only a single mysqld process is running
+	echo "Shutting down temporary server process"
+	server_pid="$1"
+	kill "$server_pid"
+	for i in $(seq 1 60); do
+		sleep 1
+		if ! $(pidof /usr/sbin/mysqld >/dev/null 2>&1); then
+			echo "Server stopped"
+			return 0
+		fi
+	done
+	# The server hasn't shut down in a timely manner
+	echo "Error: Unable to shut down server with process id $server_pid" >&2
+	return 1
+}
+_upgrade_data() {
+	local tmpdir=$1
+	chown mysql:mysql "$tmpdir"
+	# We use skip-grant-tables since we can't know the root credentials of existing databases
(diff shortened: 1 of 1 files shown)


8.0/docker-entrypoint.sh
```diff
@@ -64,6 +64,36 @@ _get_config() {
...
+	done
+	# The server hasn't shut down in a timely manner
+	echo "Error: Unable to shut down server with process id $server_pid" >&2
+	return 1
+}
+_upgrade_data() {
+	local tmpdir=$1
+	chown mysql:mysql "$tmpdir"
+	# We use skip-grant-tables since we can't know the root credentials of existing databases
+	mysqld --skip-networking --pid-file="$tmpdir/mysqld.pid" --socket="$tmpdir/mysqld.sock" --daemonize --skip-grant-tables --log-error="$tmpdir/error.log"
```
tianon: This needs to use the same `"$@"` logic we use for the other place we start the daemon, doesn't it? (which is slightly complicated by it being a shell function and `"$@"` from the script itself being obscured)

Maybe we need to _fully_ resolve what #402 starts (by functionalizing more parts of this file) so we can invoke the daemon in the same way (with a way to add `--skip-grant-tables` for this use case)?

ltangvald: Yeah, I agree. I'll do some work on that first

hachh:
Hi,
I also need this feature, is it still under development ?

TimWolla:
This PR is no longer required as of MySQL 8.0.16: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-16.html

> Previously, after installation of a new version of MySQL, the MySQL server automatically upgrades the data dictionary tables at the next startup, after which the DBA is expected to invoke mysql_upgrade manually to upgrade the system tables in the mysql schema, as well as objects in other schemas such as the sys schema and user schemas.

> The server now performs the tasks previously handled by mysql_upgrade. After installation of a new MySQL version, the server now automatically performs all necessary upgrade tasks at the next startup and is not dependent on the DBA invoking mysql_upgrade. In addition, the server updates the contents of the help tables (something mysql_upgrade did not do). A new --upgrade server option provides control over how the server performs automatic data dictionary and server upgrade operations. For more information, see Upgrading MySQL.

yosifkit:
For this image, the upgrade process is basically identical to the MySQL docs for upgrading the MySQL Server container:
https://dev.mysql.com/doc/refman/8.0/en/docker-mysql-getting-started.html#docker-upgrading

And since `8.0.16` does the upgrade automatically, I think we can close this PR.

PR #423, opened 2018-05-06
Add MYSQL_INIT_ONLY var to exit after init

cu:
One of the things I use the mysql image for is bind-mounting a dump into `/docker-entrypoint-initdb.d/` to restore a database. The database restore happens, mysql starts up, and everything is happy. However, I usually don't want to leave the dump bind-mounted into the container permanently. Once the database has been restored, I often want to stop the container, remove the dump file, and then worry about the full deployment of the service at some later stage.

However, the current `docker-entrypoint.sh` always starts mysql after the init process. If we perform a database restore as part of the init, there isn't anything to signal when a restore is complete so that the container can be shut down. So my workaround up until now has been to also bind-mount a shell script to run after the database restore. (The "signal" can be just touching the file `/var/lib/mysql/restore.done`, for example. When the file exists, our orchestration stuff knows the database has been restored and can stop/rm the container.)

This change skirts the whole issue by providing a way to tell the container to exit once all initialization is done. If there's some other less obvious way to do the same thing, I'm all ears as I haven't seen it yet.

I will submit a PR to update the docs if consensus is that this isn't a stupid idea.

If this is accepted, I will likely propose the same feature for the mariadb and postgres images as well.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -200,6 +200,11 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			exit 1
 		fi
 
+		if [ "$MYSQL_INIT_ONLY" ]; then
+			echo 'MySQL init process done. Exiting.'
+			exit 0
+		fi
+
 		echo
 		echo 'MySQL init process done. Ready for start up.'
 		echo
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -200,6 +200,11 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			exit 1
 		fi
 
+		if [ "$MYSQL_INIT_ONLY" ]; then
+			echo 'MySQL init process done. Exiting.'
+			exit 0
+		fi
+
 		echo
 		echo 'MySQL init process done. Ready for start up.'
 		echo
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -205,6 +205,11 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			exit 1
 		fi
 
+		if [ "$MYSQL_INIT_ONLY" ]; then
+			echo 'MySQL init process done. Exiting.'
+			exit 0
+		fi
+
 		echo
 		echo 'MySQL init process done. Ready for start up.'
 		echo
--- a/8.0/docker-entrypoint.sh
+++ b/8.0/docker-entrypoint.sh
@@ -207,6 +207,11 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			exit 1
 		fi
 
+		if [ "$MYSQL_INIT_ONLY" ]; then
+			echo 'MySQL init process done. Exiting.'
+			exit 0
+		fi
+
 		echo
 		echo 'MySQL init process done. Ready for start up.'
 		echo


colinmollenhour:
👍 Although a better name might be "MYSQL_EXIT_AFTER_INIT"

yosifkit:
Sorry for the long tail here; we were kind of stalled with #471. With it merged, this kind of customization can be done via a simple entrypoint script. I made this by looking at [`_main`](https://github.com/docker-library/mysql/blob/75f81c8e20e5085422155c48a50d99321212bf6f/.template.Debian/docker-entrypoint.sh#L319-L364) to see what steps I want it to perform and where I want it to stop. So, to do what this PR was doing, would be to drop the `exec` at the end (and any other parts that aren't used).

```bash
#!/usr/bin/env bash
set -Eeo pipefail

source "$(which docker-entrypoint.sh)"

### assumption: "$@" starts with mysqld followed by options (see docker run below)

mysql_note "Entrypoint script for MySQL Server ${MYSQL_VERSION} started."

mysql_check_config "$@"
# Load various environment variables
docker_setup_env "$@"
docker_create_db_directories

##### this section might be problematic if the custom script isn't designed to be restarted from the beginning
## it will just work fine with this short script but may cause issues if you have more setup above
# If container is started as root user, restart as dedicated mysql user
if [ "$(id -u)" = "0" ]; then
	mysql_note "Switching to dedicated user 'mysql'"
fi
#####

# there's no database, so it needs to be initialized
if [ -z "$DATABASE_ALREADY_EXISTS" ]; then
	docker_verify_minimum_env
	docker_init_database_dir "$@"

	mysql_note "Starting temporary server"
	docker_temp_server_start "$@"
	mysql_note "Temporary server started."

	docker_setup_db
	docker_process_init_files /docker-entrypoint-initdb.d/*

	mysql_expire_root_user

	mysql_note "Stopping temporary server"
	docker_temp_server_stop
	mysql_note "Temporary server stopped"

	echo
	mysql_note "MySQL init process done. Ready for start up."
	echo
fi
```
 Setup and then just run it (adding in your `/docker-entrypoint-initdb.d/` restore scripts):
```console
$ ls -ln
total 8
drwxrwsr-x 6 1000 1000 4096 Dec  5 14:44 data
-rwxrwxr-x 1 1000 1000 1458 Dec  5 14:39 init-only-entry.sh
$ ls -lna data/
total 165912
drwxrwsr-x 6 1000 1000     4096 Dec  5 14:44  .
drwxrwsr-x 3 1000 1000     4096 Dec  5 14:39  ..
$ docker run -it --rm \
 -e MYSQL_ROOT_PASSWORD=12345 \
 -v "$PWD/init-only-entry.sh":/usr/local/bin/init-only-entry.sh \
 -v "$PWD/data/":/var/lib/mysql \
 --name=pg \
 mysql:8.0 init-only-entry.sh mysqld
... (mysql doing init stuff and exiting)
$ $ ls -lna data/
total 165912
drwxrwsr-x 6  999 1000     4096 Dec  5 14:44  .
drwxrwsr-x 3 1000 1000     4096 Dec  5 14:39  ..
-rw-r----- 1  999 1000     1272 Dec  5 14:44  ab6f7df075c6.err
-rw-r----- 1  999 1000       56 Dec  5 14:44  auto.cnf
-rw-r----- 1  999 1000  3084516 Dec  5 14:44  binlog.000001
-rw-r----- 1  999 1000       16 Dec  5 14:44  binlog.index
-rw------- 1  999 1000     1680 Dec  5 14:44  ca-key.pem
-rw-r--r-- 1  999 1000     1112 Dec  5 14:44  ca.pem
-rw-r--r-- 1  999 1000     1112 Dec  5 14:44  client-cert.pem
-rw------- 1  999 1000     1680 Dec  5 14:44  client-key.pem
-rw-r----- 1  999 1000     5416 Dec  5 14:44  ib_buffer_pool
-rw-r----- 1  999 1000 12582912 Dec  5 14:44  ibdata1
-rw-r----- 1  999 1000 50331648 Dec  5 14:44  ib_logfile0
-rw-r----- 1  999 1000 50331648 Dec  5 14:44  ib_logfile1
drwxr-s--- 2  999 1000     4096 Dec  5 14:44 '#innodb_temp'
drwxr-s--- 2  999 1000     4096 Dec  5 14:44  mysql
-rw-r----- 1  999 1000 30408704 Dec  5 14:44  mysql.ibd
drwxr-s--- 2  999 1000     4096 Dec  5 14:44  performance_schema
-rw------- 1  999 1000     1676 Dec  5 14:44  private_key.pem
-rw-r--r-- 1  999 1000      452 Dec  5 14:44  public_key.pem
-rw-r--r-- 1  999 1000     1112 Dec  5 14:44  server-cert.pem
-rw------- 1  999 1000     1676 Dec  5 14:44  server-key.pem
drwxr-s--- 2  999 1000     4096 Dec  5 14:44  sys
-rw-r----- 1  999 1000 12582912 Dec  5 14:44  undo_001
-rw-r----- 1  999 1000 10485760 Dec  5 14:44  undo_002
```

PR #429, opened 2018-05-18
seed scripts not readable by effective user mysql when mounted volume

qrkourier:
also `chown -R .../initdb.d` to user mysql while root prior to `gosu exec mysql`

### DIFF
--- a/5.7/Dockerfile
+++ b/5.7/Dockerfile
@@ -72,7 +72,7 @@ VOLUME /var/lib/mysql
 
 WORKDIR /usr/local/bin
 COPY docker-entrypoint.sh /usr/local/bin/
-RUN ln -s usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
+RUN ln -s /usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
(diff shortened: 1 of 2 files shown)


tianon:
Sorry for the delay!

This PR is changing a number of seemingly unrelated things -- can you please describe the problem you're having (and trying to solve here) in more detail?

qrkourier:
The failure mode is that seed scripts in /docker-entrypoint-initdb.d are not readable by effective user "mysql". This problem presents when that directory is a Docker volume mount point and the mounted files are not readable by "mysql" user. The proposed solution is to change the entrypoint script to first copy the same seed scripts into the running container, set filemodes to make readable by "mysql" user before dropping privileges with gosu.

Unrelated is the problem of sporadic failure during `docker build` because the HKP keyserver is not available. The proposed solution is a list of alternative keyservers. This could easily become a separate PR if it's muddying the water.

yosifkit:
The failure mode is made clear by https://github.com/docker-library/mysql/pull/453.  As far as copying and chowning the files from `docker-entrypoint-initdb.d` I would say no since some users already create an image with their files built in (instead of bind-mounting) and that this would duplicate the required space which might be a full database dump of multiple gigabytes.

PR #438, opened 2018-06-06
Annotate potential chown-mistake

MikesCodeProjects:
Mounting an nfs-Folder to /var/lib/mysql may causes problems, starting container successfully.
Container exits when trying to set owner and group as mysql user through docker-entrypoint.sh.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -71,7 +71,7 @@ RUN mkdir -p /etc/mysql/conf.d \
 	} > /etc/mysql/my.cnf
 
 RUN mkdir -p /var/lib/mysql /var/run/mysqld \
-	&& chown -R mysql:mysql /var/lib/mysql /var/run/mysqld \
+	#&& chown -R mysql:mysql /var/lib/mysql /var/run/mysqld \
 # ensure that /var/run/mysqld (used for socket and lock files) is writable regardless of the UID our mysqld instance ends up having at runtime
 	&& chmod 777 /var/run/mysqld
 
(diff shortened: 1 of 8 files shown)


tianon:
Uh, this definitely wasn't a mistake (and commenting it out isn't the solution).  I think you'll need to elaborate a bit more on the specific problem you're facing?

Have you tried using a subdirectory on NFS instead of using the root? (which is always a good idea, regardless of the underlying filesystem when bind-mounting into a container)

Have you tried using `--user` to both skip this code and force the container to run as a UID/GID that have explicit access to the NFS share you're trying to use?

MikesCodeProjects:
**An application will always** be able to read from a file and write to it **if it has the permissions** to do that.
There is no need to change the owners and make the startup success depend on it.

**Using a Network-File-System means,**
both systems (the nfs-server and the nfs-client) will use different user-databases except you'r using a centralized system or map them somehow.
Since I am using one user on the nfs-server and two users on the nfs-client, I cant map them.

This pull-request only means: It is better to remove the chown-statement because it is only another conditional-dependency which can lead to an unsuccessful start.

MikesCodeProjects:
I mean... It's been a while. You can actually merge this commit or just keep depending on chown-ing files as 'mysql' (non-root) user.

yosifkit:
Yeah, this is not something we would merge since it breaks when using a bind mount (`-v /local/path/:/var/lib/mysql/`). If you need to prevent the `chown` then use `--user` with a numeric ID on the `docker run` (or the equivalent compose file field).

PR #449, opened 2018-07-09
docker image size improvement

valerianpereira:
docker image size improvement -
every RUN command adds a layer to the image.


tianon:
Layer count and image size are very different concepts -- decreasing the number of layers will not necessarily decrease their size (or the overall image size), and in this case I think we gain more in readability with separate layers than combining gains us (since this won't actually make the final image any smaller).

PR #453, opened 2018-07-16
fix silently skipped init scripts

lalyos:
The for loop calling process_init_file in [docker-entrypoint.sh](https://github.com/docker-library/mysql/blob/master/8.0/docker-entrypoint.sh#L196) silently skips if there are permission issues.

## Initializing

the official docs abour [initializing-a-fresh-instance](https://github.com/docker-library/docs/tree/master/mysql#initializing-a-fresh-instance) instructs to bind mount the **/docker-entrypoint-initdb.d/** dir:

```
docker run \
  -d \
  --name db \
  -e MYSQL_DATABASE=mydb \
  -e MYSQL_ROOT_PASSWORD=secret \
  -v $PWD:/docker-entrypoint-initdb.d/ \
  mysq
```

the `for` cycle on a non readable directory silently skips the loop body, leaving the user wonder what's wrong with scripts. 

## Proposed fix

by adding an extra `ls` command would produce an error message:
```
ls: cannot open directory '/docker-entrypoint-initdb.d/': Permission denied
```


lalyos:
Hi, @tianon, the purpose of `[ -e /docker-entrypoint-initdb.d/ ]` is to only try to `ls` if the directory exists, otherwise every container, which doesn't bind anything to /docker-entrypoint-initdb.d/ would fail.

stdout of `ls` is redirected to `/dev/null`

tianon:
Right, that's what's got me confused -- it'd be pretty unique for someone to not have that directory existing, given that it's created in the base:

https://github.com/docker-library/mysql/blob/075c24deda2f551531c76ef46283c8c9c810d78f/8.0/Dockerfile#L23

tianon:
(And in that case, I think it's totally appropriate for folks to get a harmless `ls` error.)

lalyos:
Ohh i see, that makes more sense. I just removed the extra check.

tianon:
Once Travis is happy again I think we're good to merge. :+1:

Issue #410, opened 2018-04-20
mysqladmin and other tools missing from mysql:8


stealthybox:
Hi 👋 

The recent tag move for `latest` -> `8` started failing our kubernetes liveness probes:

`Liveness:   exec [mysqladmin ping --host 127.0.0.1 --port 3306] delay=5s timeout=5s period=10s #success=1 #failure=3`

The mysql docs suggest that this tool should be usable with both MySQL 5.7 and 8.0:
- https://dev.mysql.com/doc/refman/5.7/en/mysqladmin.html
- https://dev.mysql.com/doc/refman/8.0/en/mysqladmin.html

It seems there may be other tools missing from the `mysql:8` image as well:
```bash
comm \
<(docker run --entrypoint sh -w /usr/bin mysql:5 -c 'ls mysql*') \
<(docker run --entrypoint sh -w /usr/bin mysql:8 -c 'ls mysql*')

		mysql
mysql_config_editor
mysql_embedded
mysql_install_db
mysql_plugin
		mysql_secure_installation
		mysql_ssl_rsa_setup
		mysql_tzinfo_to_sql
		mysql_upgrade
mysqladmin
		mysqlbinlog
mysqlcheck
		mysqld_multi
		mysqld_safe
		mysqldump
mysqldumpslow
mysqlimport
		mysqlpump
mysqlshow
mysqlslap
```

Cheers

ltangvald:
Hi,

This is an oversight in our server-core packages, sorry. The contents of the package are listed in https://github.com/mysql/mysql-server/blob/8.0/packaging/deb-in/mysql-packagesource-server-core.install.in

We did drop some of the executables from the core package, since we don't consider them necessary for most people, but mysqladmin should be there (it's in the minimal rpm used for the upstream docker image)

stealthybox:
@ltangvald thanks for the reference.

Will `mysqladmin` be added back into the debian package?
I currently don't see it in that install file unless I'm missing something implicit.

tianon:
Essentially https://github.com/mysql/mysql-server/commit/625609bbb9034007f498d40d5c337e9bc3a2ae87, but to add `mysqladmin`? :sweat_smile:

stealthybox:
@tianon aha! yeah 😅

Coolfeather2:
got same problem, waiting for new release

falcoschwarz:
Please also add back mysql_config_editor. It was such an elegant way to not store any passwords in clear text.

[...]

PR #465, opened 2018-08-01
Install mysql-community-client instead of mysql-community-client-core

ltangvald:
The core package is missing certain utilities users may want, and using it doesn't save all that much in image size compared to just installing the full client (~40MB in unpacked image size)

Resolves issue #410


tianon:
Doh, should've merged this sooner, sorry!!  LGTM :+1:

PR #466, opened 2018-08-01
Expose mysqlx port

gpinhey:
The mysql v8 x-plugin [listens on port 33060](https://dev.mysql.com/doc/refman/8.0/en/x-plugin-system-variables-options.html#option_mysqld_mysqlx-port) by default, so this port should be marked as exposed.


tianon:
The upstream-maintained image does `EXPOSE` this port as well, so I suppose I'm +1:

https://github.com/mysql/mysql-docker/blob/8e8a22d1f78c47ac6d80ba0636de04fa34116e55/8.0/Dockerfile#L32

@ltangvald is this sane for us to enable here too?

ltangvald:
Yeah, I think this is good to have (though from what I understand, EXPOSE doesn't actually publish the port unless -p or similar is also used?)

ltangvald:
One question, though: Does it make more sense to use EXPOSE 3306 33060 instead of having it on multiple lines? Is there a recommended standard for this?

gpinhey:
Since the company-official mysql image puts both ports on one line, I'll change my PR to do the same.

It looks like `EXPOSE` in the Dockerfile is primarily used for inter-container communication where multiple containers exist on the same network, you don't need to `EXPOSE` a port if you'll be using `docker run -p` to explicitly bind a port from docker to host. However, in those inter-container communication environments (whether by `docker-compose` or, in my case, CircleCI) it helps to have the ports already open.

Plus, using `EXPOSE` is built-in documentation for what ports your docker expects to use.

gpinhey:
Upstream image and docs suggest this should be opened in 5.7 as well:
https://github.com/mysql/mysql-docker/blob/8e8a22d1f78c47ac6d80ba0636de04fa34116e55/5.7/Dockerfile#L32
https://dev.mysql.com/doc/refman/5.7/en/x-plugin-system-variables-options.html#option_mysqld_mysqlx-port

It ends there, as 5.6 only opens the standard 3306 port:
https://github.com/mysql/mysql-docker/blob/8e8a22d1f78c47ac6d80ba0636de04fa34116e55/5.6/Dockerfile#L32

tianon:
> Does it make more sense to use EXPOSE 3306 33060 instead of having it on multiple lines? Is there a recommended standard for this?

It doesn't really make a difference (except for personal preference reading the `Dockerfile`); either way, Docker will collapse both together into a single metadata change, so they're functionally identical.

This LGTM either way.

Issue #357, opened 2017-12-21
Add a feature to shtudown container after running init scripts


mjpowersjr:
Sometimes I use a command like the following to initialize a mysql new database volume from a backup.
```bash
docker run --rm ... --volume /path/to/backup.sql:/docker-entrypoint-initdb.d/backup.sql mysql
```

It would be handy if I could (through a environment variable, alternative entrypoint, etc.) cause the docker container to stop running after the init process is complete, instead of listening for connections.

My current workaround (hack) is to build a new image based on the mysql image, that contains a shell script that waits for the init process to finish - and then kill the mysql service. This seems to work fine, but it would be nice if there was a baked in way to accomplish this behavior.
```bash
#!/bin/bash

set -e

# wait until init process is finished
until /docker-entrypoint.sh mysqld 2>&1 | \
    tee /dev/tty | \
    grep --silent --extended-regexp --max-count=1 "init process done|error";
do 
    sleep 1; 
done

# shutdown container
kill -s TERM $$
```

yosifkit:
Why not just make your script alphabetically after your backup file in `docker-entrypoint-initdb.d` and then you can do something like the following:

```bash
#!/bin/bash
set -e

# you should have access to any vars from the entrypoint since scripts are sourced
kill -s TERM "$pid"
wait "$pid"
# or
killall -s TERM mysqld
sleep 10

# prevent entrypoint from continuing
exit 1
```

davidroeca:
The solution documented in this issue no longer works (`$pid` is no longer an accessible variable). The solution mentioned in https://github.com/docker-library/mysql/pull/423#issuecomment-562355462 works but will have to be manually updated every time the implementation changes.

Are there plans to support this now that https://github.com/docker-library/mysql/pull/471 is merged?

tianon:
You should be able to invoke `docker_temp_server_stop` to stop the temporary server, followed by `exit` in order to stop the execution of any further instructions (as above).

tianon:
Also, with #471 you could more easily write your own script which just does the initialization followed by exiting (and thus be much less error-prone).

davidroeca:
A script that copies `_main` from `docker-entrypoint.sh` becomes error-prone the moment a name used within `_main` changes, or a new (and necessary) command is added. What I'd like to do is just run `_main` without the last line.

davidroeca:
Maybe I'm confused and there's an easier way to go about this?

mjpowersjr:
[...]

Issue #318, opened 2017-08-16
support arm64 architecture


longquan7:
I want to build this image on arm64 architecture but I don't find arm64 package in this file http://repo.mysql.com/apt/debian/dists/jessie/InRelease , So I want to know the image whether or not support arm64 architecture.

yosifkit:
This would need support from MySQL upstream to release packages for other architectures in their apt repo. (to me, this seems unlikely since they don't commercially support much more then i386 and amd64: https://www.mysql.com/support/supportedplatforms/database.html)

cc @ltangvald would mysql be willing to build their apt packages on more architectures? (like arm64v8, arm32v7, ppc64le, and s390x)

gnumoksha:
@yosifkit  [debian has mysql server package](https://packages.debian.org/search?suite=stretch&arch=arm&searchon=names&keywords=mysql) for arm architecture. I am using it on a few Raspberry Pi.

tianon:
Debian's versions won't be up to date with upstream's, which is why these
images consume packages exclusively from upstream's repositories.

In the meantime, I'd recommend creating your own image which installs
Debian's packages if this is something you need right away.

smallfryy:
@longquan7 - did you find a solution? I'm also looking to build mySQL:5.7 image on arm64 architecture (Odroid XU3, per issue referenced above).

longquan7:
@samalba Nothing yet.This would need support from MySQL upstream to release packages.

geerlingguy:
Just a note for anyone who needs a working image for testing purposes (though I'd recommend building your own for prod!) for Raspberry Pi at least—I've tested the following two and can confirm they work okay on my Pi 3 B and 3 B+ with Raspbian:

  - https://hub.docker.com/r/hypriot/rpi-mysql/
  - https://hub.docker.com/r/tobi312/rpi-mysql/

Note that they only support MySQL 5.5 at this time.

shawnguo2:
@ltangvald I'm particularly wondering about mysql's take on arm64 architecture support, considering there are a lot of push on arm64 enterprise from industry.  Is there any plan already being discussed or just no, nothing on that?  Thanks much.

ltangvald:
We do have an experimental preview out https://blogs.oracle.com/wim/mysql-805-community-edition-preview-for-oracle-linux-7-for-arm64-preview (8.0.5 was never released, but was instead renamed to 8.0.11), but I don't really have any more concrete information than that :)

[...]

PR #471, opened 2018-08-07
Templatize and refactor 5.7+ entrypoint scripts

### LINKED ISSUE #490: added .cnf affects image initialization
If mysql `.cnf` is added when container starts, and it has a `[client]` section, incorrect password will be used in the init script.

```
[client]
user=root
password=unprotected
database=xxx_development
```

then init will fail like this:

```
mysql_1                  | MySQL init process in progress...
mysql_1                  | 2018-09-13T02:53:44.017882Z 29 [Note] Access denied for user 'root'@'localhost' (using password: YES)
mysql_1                  | MySQL init process in progress...
mysql_1                  | 2018-09-13T02:53:45.025505Z 30 [Note] Access denied for user 'root'@'localhost' (using password: YES)
mysql_1                  | MySQL init process in progress...
mysql_1                  | 2018-09-13T02:53:46.032175Z 31 [Note] Access denied for user 'root'@'localhost' (using password: YES)
mysql_1                  | MySQL init process in progress...
mysql_1                  | MySQL init process failed.
```

We should probably explicitly use password here:
https://github.com/docker-library/mysql/blob/b39f1e5e4ec82dc8039cecc91dbf34f6c9ae5fb0/5.7/docker-entrypoint.sh#L123

ltangvald:
Note: May want to split this up, but putting everything in a single request for any discussions that may arise.

Unifies the entrypoint scripts for 5.7 and 8.0, since they do not need to be different at all, and adds them to a debian template that is copied into each directory by update.sh

Also refactors the entrypoint script by moving logic for starting and stopping the server into functions and adding timestamps and error levels to the echo lines.

Closes #490


update.sh
```diff
@@ -14,7 +14,14 @@ declare -A debianVariants=(
...
 )
 
+# Copy entrypoint template
+templateVersions=( "5.7 8.0" )
+for version in ${templateVersions}; do
+	cp "template.Debian/docker-entrypoint.sh" "${version}/"
+done
+
 for version in "${versions[@]}"; do
+	if [ "${version}" = "template.Debian" ]; then continue; fi # If update.sh is run without arguments, the template directory is included in the list
```
tianon: What if we instead name the directory starting with a `.` so that it's "hidden" and just naturally gets excluded from this loop?

update.sh
```diff
@@ -14,7 +14,14 @@ declare -A debianVariants=(
 	#[5.5]='jessie'
 )
 
+# Copy entrypoint template
+templateVersions=( "5.7 8.0" )
+for version in ${templateVersions}; do
+	cp "template.Debian/docker-entrypoint.sh" "${version}/"
```
tianon: There's really not a _whole_ lot that's different between these and 5.5/5.6 -- do you think it'd make sense to do some runtime detection and switch behavior in the scripts dynamically in order to hit all four versions?  We could also have this script include the nuances for the lower versions and have it dynamically update the template when copying it to the lower versions.

template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,238 @@
...
+# usage: file_env VAR [DEFAULT]
+#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
+# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
+#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
+file_env() {
+	local var="$1"
+	local fileVar="${var}_FILE"
+	local def="${2:-}"
+	if [ "${!var:-}" ] && [ "${!fileVar:-}" ]; then
+		echo >&2 "$(date --rfc-3339=seconds) [ERROR] [Entrypoint]: Both $var and $fileVar are set (but are exclusive)"
```
tianon: This all seems to use `echo [>&2] "$(date --rfc-3339=seconds) [ERROR/Note] [Entrypoint]: ...` quite a lot; do you think it'd make sense to make some simple helpers out of that so we stay consistent?  Something like:

```bash
_log() {
	echo "$(date --rfc-3339=seconds) [$1] [Entrypoint]: $*"
}
_note() {
	_log Note "$@"
}
_error() {
	_log ERROR "$@" >&2
}
```

template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,238 @@
...
+	fi
+	return $pid
+}
+
+_stop_server() {
+	local server_pid="$1"
+	kill "$server_pid"
+	for i in $(seq 1 60); do
+		sleep 1
+		if ! $(pidof /usr/sbin/mysqld >/dev/null 2>&1); then
```
tianon: Using `pidof` seems kind of hacky, especially if someone were to run this image with `--pid=host`; couldn't we instead use `--pid-file` (https://dev.mysql.com/doc/refman/8.0/en/mysql-server.html#option_mysql_server_pid-file) to have `mysqld` itself store this in a useful place for us to look up?

(Given that it has a reasonable default and doesn't appear to be disable-able, we might be able to get away with something like `_get_config 'pid-file'` to capture the appropriate value, but I'm not 100% sure where it'd be relative to if that is configured with a relative path?)

template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,238 @@
...
+		if echo 'SELECT 1' | "${mysql[@]}" &> /dev/null; then
+			break
+		fi
+		sleep 1
+	done
+	if [ "$i" = 0 ]; then
+		echo >&2 "$(date --rfc-3339=seconds) [ERROR] [Entrypoint]: Unable to start server."
+		exit 1
+	fi
+	return $pid
```
tianon: Won't this cause anything in this function to be interpreted as the PID, even if it's a non-zero exit code due to `set -e`? (see other comment about potentially using `--pid-file` instead)

For example, if the initial `mysqld` invocation fails outright, `set -e` will cause this function to exit with a non-zero exit code (instead of going through the `for` loop and getting to the `exit 1` line which would stop the whole script), and that random non-zero exit code would be returned to the function caller who expects it to be a process ID (and that `mysqld` is actually running, when it isn't).

template.Debian/docker-entrypoint.sh
ltangvald: Hm, yes. Normally just using the pid-file isn't enough (since it may have stale data, leading you to killing some other process instead).
And pid-file should never be a relative path :)

A couple of notes about starting/stopping the server, that we've implemented upstream and I might add to this PR:
1. 5.7+ Has the --daemonize option, which forks the process when ready, meaning no need to have a loop that checks if the server is up (https://github.com/mysql/mysql-docker/blob/mysql-server/5.7/docker-entrypoint.sh#L79)
2. When using a local socket, «mysqladmin shutdown» will block until the server has shut down (for 5.5+), so as long as the script has root access to the server, which it will during a fresh start (https://github.com/mysql/mysql-docker/blob/mysql-server/5.7/docker-entrypoint.sh#L160).

1. Is only a help for 5.7, but 2. should remove the need to pass the pid around

template.Debian/docker-entrypoint.sh
ltangvald: Yes, I think you're right (see above comment for a possibly better way to handle this)

update.sh
ltangvald: Good point! Will make that change

update.sh
ltangvald: Yeah, you're right it's not that much work (we do this at https://github.com/mysql/mysql-docker/blob/mysql-server/gen_dockerfiles.sh).

template.Debian/docker-entrypoint.sh
ltangvald: Yes, definitely. Will add it :)

template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,238 @@
...
+}
+
+_start_server() {
+	local socket=$1; shift
+	"$@" --skip-networking --socket="${socket}" &
+	local pid="$!"
+
+	mysql=( mysql --protocol=socket -uroot -hlocalhost --socket="${socket}" )
+
+	for i in {30..0}; do
```
yosifkit: Should we address #159 and #288 and add `MYSQL_START_TIMEOUT` or similar name here?

ltangvald: If we use --daemonize we shouldn't need a timeout at all; The call would block until the server is ready, quickly returning a nonzero code if it fails and otherwise returning 0 when the server is ready for use. We could maybe add a wait on it just in case something could cause server startup to hang indefinitely, but I'm not aware of any cases like that (and it adds complexity to the script)

yosifkit: Just using a `daemonize` that waits until ready sounds better and is basically what we use for [postgres](https://github.com/docker-library/postgres/blob/35d6f4b68399cf2ffcb1c13c27003b92f53b9bc8/docker-entrypoint.sh#L114-L116).

tianon:
@ltangvald is there anything @yosifkit and I can do to help you move this forward and/or support you better? :smile:

ltangvald:
Right now it's hanging on me, so only if you can find some more hours in the day :D
But I have some time this week, so will update the PR with the suggestions we've discussed and push it :)

template.Debian/docker-entrypoint.sh
ltangvald: Also added a _warn (used it for the notice about files in the initdb.d directory being ignored)

ltangvald:
Commits should resolve all conversations except the one about adding 5.5 and 5.6 to the template. We have this upstream, but the implementation results in chunks of dead code in the entrypoint scripts of the newer versions: https://github.com/mysql/mysql-docker/blob/mysql-server/5.7/docker-entrypoint.sh#L89.
Do you have any ideas for how to better resolve this? The problem is mostly just the non-daemonize method for starting the server.

I also added one change for storing the root password in a file instead of on the command line, since mysqladmin also needs it.

ltangvald:
@tianon: Sorry, yes, it was meant to be review-ready :)
However, I also added a "docker" prefix to the function names now (from the discussion at https://github.com/docker-library/postgres/pull/496) plus a small typo fix.

We could add the other issues, but they could just as well be done after, and the templating should make future fixes easier to handle

ltangvald:
@yosifkit Good if you also take a look at the changes :)

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,259 @@
...
+fi
+
+if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
+	# still need to check config, container may have started with --user
+	docker_check_config "$@"
+	# Get config
+	DATADIR="$(docker_get_config 'datadir' "$@")"
+
+	if [ ! -d "$DATADIR/mysql" ]; then
+		docker_file_env 'MYSQL_ROOT_PASSWORD'
```
glensc: why not move this block to subroutine as well?

yosifkit:
My thoughts are along the lines that maybe this doesn't go far enough with functionalizing the entrypoint.  This could be separate discussion, but I think if we are going to create functions in the entrypoint that users can use from initidb.d scripts, then we might as well go all the way to providing a consistent well-reasoned interface.

If https://github.com/docker-library/postgres/pull/496 seems reasonable, I'd like to get MySQL (and mariadb + percona) as similar as possible.  I am definitely open to suggestions for improvement there; especially if it will make it work better for MySQL.  This would allow users more places to hook in customization by sourcing the entrypoint script with their own script without us having to explicitly support more environment variables, an always run initidb.d directory (https://github.com/docker-library/percona/issues/58), or an only run initdb.d (https://github.com/docker-library/mysql/pull/357, https://github.com/docker-library/mysql/issues/423).

Where can we collaborate to bring to two ideas closer together? Would defining the function list (and possible arguments) be a good place to start?  Here is what i have so far in the postgres rework: 

```bash
file_env
docker_create_database_dirs
docker_init_database_dir
docker_print_password_warning
docker_process_init_files
# maybe drop the `p`, or use docker_process_sql or something
docker_psql_run
docker_setup_database
docker_setup_env_vars
# there isn't really a MySQL equivalent of this one, so it should probably be unique to postgres
docker_setup_pg_hba
# perhaps these should be a more generic "docker_temporary_sqlserver_start"
docker_temporary_pgserver_start
docker_temporary_pgserver_stop
```

tianon:
Friendly ping @ltangvald :heart:

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,259 @@
...
+			# https://github.com/mysql/mysql-server/blob/23032807537d8dd8ee4ec1c4d40f0633cd4e12f9/packaging/deb-in/extra/mysql-systemd-start#L81-L84
+			docker_note "Initializing certificates"
+			mysql_ssl_rsa_setup --datadir="$DATADIR"
+			docker_note "Certificates initialized"
+		fi
+
+		SOCKET="$(docker_get_config 'socket' "$@")"
+		# We create a file to store the root password in so we don''t use it on the command line
+		install -d -m0700 /tmp/mysql-files
+		PASSFILE=$(mktemp /tmp/mysql-files/XXXXXXXXXX)
```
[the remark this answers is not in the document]
ltangvald: Right. The use of install was to ensure the correct file permissions, but they are default from mktemp anyway.

ltangvald:
@tianon Working on moving as much logic as possible into functions, as yosifkit suggests :)

ltangvald:
@tianon @yosifkit I've refactored further to put as much logic as possible into functions. The list is a bit different, but could you look it over and maybe suggest refinements to bring it closer to pg?

glensc:
@ltangvald there's still rather large block in global scope: https://github.com/docker-library/mysql/blob/94616c4269fda0bfac15894be34d2fb6215b4404/.template.Debian/docker-entrypoint.sh#L270-L318

why not move that into function as well?

ltangvald:
That block is (almost) the entire script; Overriding it essentially means overriding the entire entrypoint.
But maybe it would be useful to move all the helper functions into a separate file so they would be available for someone who wants to define their own behavior from scratch.

glensc:
imho the only goal should not be override possibility, but code readability too. moving to function is more compact and allows using local variables. 

with that in mind, i tend to write my shell scripts nowadays as:

```
#!/bin/sh

func() {
  ...
}

main() {
  local foo=$1 bar
  func bar
}

main "$@"
```

- [1](https://github.com/glensc/docker-create-tag/blob/3e62cb0bdea64e28a0f48e95d67a4a50730733cc/docker-create-tag.sh)

ltangvald:
Added a docker_main function now (there was also a bug with non-mysqld commands failing because it would still try to fetch server config for them).

About local functions: I'm on pretty shaky ground with bash scoping, but currently the script relies on global variables for certain settings like DATADIR and SOCKET

ltangvald:
Added a docker_main function now (there was also a bug with non-mysqld commands failing because it would still try to fetch server config for them).

About local variables: I'm on pretty shaky ground with bash scoping, but currently the script relies on global variables for certain settings like DATADIR and SOCKET

yosifkit:
I thought about moving the helper functions to another file for postgres, but instead I just hid the main logic in a function and only run that function when the file is not sourced so that a user can source our one file and then choose which pieces they want to run.  That way we don't have to worry about telling users about a second file that they can source to get our init steps.

```bash
# check to see if this file is being run or sourced from another script
￼ _is_sourced() {
￼ 	# https://unix.stackexchange.com/a/215279
￼	[ "${FUNCNAME[${#FUNCNAME[@]} - 1]}" == 'source' ]
￼}
_main() {
	# init logic
}

if !_is_sourced; then
	_main "$@"
}
```

Do you think this would be good to do in the MySQL image too?

As for function names and usage, i'm still lining up the MySQL ones with the ones I proposed for PostgreSQL Server so that we can get a unified set (plus custom ones for each).

tianon:
> I think you're supposed to use `[[`'s with `==`, or if you want to be POSIX compatible then `[`'s go with single `=`.

It should really drop the `==` down to just `=` -- using `[[` with `==` is a Bashism (and is only necessary for glob matching, which is more reliably performed via `case xxx in GLOB) ...` instead).

(But if we're getting that picky, `FUNCNAME` isn't acceptable since it's also a Bashism :laughing:)

ltangvald:
I don't think you can be _too_ picky with bash :P 
I'll add that check.

glensc:
the shebang is bash in this script: `#!/bin/bash` so it's ok to use bashism, just care must be taken that the feature is not too fresh. but as the source environment is controlled (base os is picked by docker-library), it's irrelevant as well.

i mean the script does not need to be portable if you control the environment. there's nothing bad in being a bash script, just don't advertise as /bin/sh when you do.

that's my opinion on the bashism vs being posix compatible.

ltangvald:
Good point. [ ! var = val ] is used in  some places and var != val in others. I suspect I just took it from the [ ! -z .. ] checks. Will change them all to != to be more consistent.

ltangvald:
@glensc ^

lag-linaro:
Since this is currently blocking https://github.com/docker-library/mysql/issues/318, it would be really good to have it reviewed/merged. :smiley:

glensc:
LGTM!

jainprashantk:
Shall we?

lag-linaro:
Do it, do it, do it! :smile:

yosifkit:
Did we want to try and unify the function names here and proposed for https://github.com/docker-library/postgres/pull/496, like I had suggested above: https://github.com/docker-library/mysql/pull/471#issuecomment-426391778?  I'd love to provide a similar interface for the two (plus mariadb) for users that want to customize their sql initialization process.

As for the content of the PR, I don't like separating the entrypoint logic into `update.sh`, and would rather just add some logic in the entrypoint (we have `MYSQL_MAJOR` variable available)

Once we have finalized the changes to the entrypoint, we would need another commit that runs `update.sh` so that the changes can be tested by travis (probably after a rebase + force push).

ltangvald:
You mentioned you were lining up the functions for MySQL with PostgreSQL. Do you have a proposal for this? I don't know PG very well, but I think database setup works a bit differently, so not sure a 1:1 mapping is possible?

Using update.sh is how the postgres image does it (it's what I based this on). And the entrypoint script is already quite complex, so I'd like to avoid including a lot of extra if-else blocks into it, especially since some of it will likely be temporary (5.5 is EOL as of December 2018).

yosifkit:
Ok, went through each of the functions and lined up the ones that provide similar functionality and added some thoughts on each (I could put this in a google sheet if that would be easier to hash through).

| mysql                           | postgres                          | proposed unified name                   | functionality                                                                                                                                    | action/comments                                                                                                                                              |
|---------------------------------|-----------------------------------|-----------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `docker_file_env`               | `file_env`                        | `file_env`                              | official-image helper that is in many entrypoint scripts                                                                                         | rename back to file_env to remain consistent with the rest of the images                                                                                     |
|                                 | `docker_create_database_dirs`     | ?                                       | used to create initial directories and if run as root, ensure ownership to the correct user                                                      | move current functionality in mysql `main` to a function?  This function is called twice if the script starts as `root`                                      |
| `docker_check_config`           |                                   | n/a                                     | sanity check to make sure config files are parsable                                                                                              | postgres analogue not needed                                                                                                                                 |
| `docker_init_database_dir`      | `docker_init_database_dir`        | `docker_init_database_dir`              | `mysql_install_db`, `--initialize-insecure` and `initdb`, postgres uses `POSTGRES_INITDB_ARGS` rather than `"$@"`                                | :+1:                                                                                                                                                         |
| `docker_verify_env`             | `docker_print_password_warning`   | `docker_verify_password_env`            | postgres server allows an empty root password, this prints a warning if it is while mysql checks for missing password env and exits if all empty | perhaps these could have the name unified even though one warns and one exits?                                                                               |
| `docker_process_init_file`      | `docker_process_init_files`       | ?                                       | function that can take a directory and runs executable `sh` scripts, sources non-executable sh scripts, and processes `sql` / `sql.gz` files     | provides `"${mysql[@]}"` or `"${psql[@]}"' for backwards compatability                                                                                       |
|                                 | `docker_psql_run`                 | `docker_sql_run` / `docker_process_sql` | function to replace the `"${mysql[@]}"` and `"${psql[@]}"` that we advertised previously                                                         |                                                                                                                                                              |
| `docker_init_client_command`    |                                   |                                         | creates variable for `"${mysql[@]}"`                                                                                                             | replace with above function                                                                                                                                  |
| `docker_write_password_file`    |                                   |                                         | creates a file with the password for DB accces                                                                                                   | probably just called from or combined into `docker_sql_run`?                                                                                                 |
|                                 | `docker_setup_database`           |                                         | create initial DB from env                                                                                                                       |                                                                                                                                                              |
| `docker_init_database_user`     |                                   |                                         | create initial DB and non-root user with access                                                                                                  | probably rename this along with the postgres `docker_setup_database` (or make two functions if creating the DB + permissions separate from user makes sense) |
| `docker_get_config`             |                                   | n/a                                     | helper function for scraping mysql config                                                                                                        | postgres analogue not needed                                                                                                                                 |
| `docker_init_env`               | `docker_setup_env_vars`           | ?                                       | gathers important environment variables (using file_env or config scraping)                                                                      | pick a good name?                                                                                                                                            |
|                                 | `docker_setup_pg_hba`             | n/a                                     | sets up pg_hba.conf                                                                                                                              | mysql analogue not needed                                                                                                                                    |
| `docker_init_root_user`         |                                   | n/a                                     | creates the initial `root` database user                                                                                                         | postgres gets this for free in `docker_init_database_dir`                                                                                                    |
| `docker_generate_root_password` |                                   |                                         | creates a random password for root                                                                                                               | perhaps just move this logic into `init_root_user`?                                                                                                          |
| `docker_expire_root_user`       |                                   | n/a                                     | expires the root password                                                                                                                        | maybe push the `if [ ! -z "$MYSQL_ONETIME_PASSWORD" ]` into the function?                                                                                    |
| `docker_load_tzinfo`            |                                   | n/a                                     | loads timezone info                                                                                                                              | postgres analogue not needed  perhaps move skip check into function?                                                                                         |
| `docker_wait_for_server`        |                                   |                                         | waits for the sql server to be ready                                                                                                             | not needed for postgres; should this just be part of `docker_temporary_sqlserver_start`?                                                                     |
| `docker_start_server`           | `docker_temporary_pgserver_start` | `docker_temporary_sqlserver_start`      | starts a local only sql server, function arguments are passed along to the server daemon as flags                                                |                                                                                                                                                              |
| `docker_stop_server`            | `docker_temporary_pgserver_stop`  | `docker_temporary_sqlserver_stop`       |                                                                                                                                                  |                                                                                                                                                              |
| `docker_main`                   | `_main`                           | ?                                       | regular setup with minimal extra logic, ends with `exec "$@"`, not recommended to be used when extending                                         | plan is to keep as much logic as possible in the functions this calls                                                                                        |
| `docker_log` (note warn error)  |                                   | n/a                                     | helpers to make entrypoint messages match mysql logging                                                                                          | no postgres analogue planned                                                                                                                                 |

ltangvald:
On the whole I think it looks pretty good, but yeah, easier to comment each one if you put it in a google sheet :)

yosifkit:
@ltangvald, sure, feel free to add lots of comments https://docs.google.com/spreadsheets/d/1yYlDjwHw58ng_9eZntI2PAuGGn_WgdatqUZQp5Ra-_s/edit?usp=sharing

yosifkit:
@ltangvald, I think we've reached consensus on the set of functions. Should I work on a more formal spec for the functions or is the spreadsheet clear enough that we can each work on the PostgreSQL and MySQL implementations?

I am glad to help on the MySQL implementation as well :wink: (and hopefully we can migrate it to the MariaDB image too!)

ltangvald:
I think the spreadsheet is enough to implement, but since part of the idea is to give users a common set to work with, I guess we need it properly documented anyway :)

ltangvald:
I think I've resolved all the points from the spreadsheet now, finally (but changed docker_setup_db_users to just docker_setup_db). A basic startup works, but needs more thorough testing to ensure I didn't break anything :)

ltangvald:
@tianon @yosifkit I've run through the functionality and looks like it works. Could you look through and see if you have any more comments (there is one issue with MYSQL_ONETIME_PASSWORD only being applied to root@%, but it's an existing problem, so might be best to add a separate bug for it).

yosifkit:
I see a few edge cases, so I'd like to come back with some suggested changes next week. Just wanted to let you know I've been looking it over and will have concrete suggestions later.

yosifkit:
Ok, made my commit over in https://github.com/infosiftr/mysql/tree/refactoring (but can push here if you think it looks good).

https://github.com/infosiftr/mysql/commit/7fc6c012ca2a542d90d268aa7f998b597937dad6:
A few entrypoint updates to increase usability

`docker-entrypoint.sh`:

 - update `docker_process_init_files` to take in a list of file
 - update `docker_process_sql` to take sql input on stdin to maintain compatibility with `mysql` (and prevent us from loading an entire `.sql` file into memory)
 - move any code that was outside a function to `_main` or other functions
 - use `local` variables in functions to limit scope
 - move current `docker_create_db_directories` logic to `docker_init_database_dir`
   - new docker_create_db_directories that just creates and chowns directories
 - move `PASSFILE` to be in memory only so that we don't accidentally not clean it up
 - move `FLUSH PRIVILEGES` to fix docker-library#549

update.sh:

 - add back `DELETE` statement for 5.5 and 5.6
 - swap some template strings to single quotes to make it easier to read
 - swap `sed` to use ! for delimter
 - use one loop rather than two

----------------

My only thought now it to possibly alphabetize/organize the functions, but that can be done another time.

glensc:
I'd really love to see this merged as I'd like to have such features:
- load .xz .sql dumps
- load sql files from multiple directories

ltangvald:
@yosifkit LGTM

yosifkit:
Did a quick rebase to get the 5.5 EOL and add an extra commit to apply update.sh so we can have the tests run :smile: :tada: :rocket:

LGTM, cc @tianon.

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+	export "$var"="$val"
+	unset "$fileVar"
+}
+
+# usage: docker_process_init_files [file [file [...]]]
+#    ie: docker_process_init_files /always-initdb.d/*
+# process initializer files, based on file extensions
+docker_process_init_files() {
+	# mysql here for backwards compatibility "${mysql[@]}"
+	mysql=( docker_process_sql )
```
tianon: This should work fine, but I want to verify: was this tested and verified to work properly with a shell function?

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+			*)        mysql_warn "$0: ignoring $f" ;;
+		esac
+		echo
+	done
+}
+
+mysql_check_config() {
+	local toRun=( "$@" --verbose --help ) errors
+	if ! errors="$("${toRun[@]}" 2>&1 >/dev/null)"; then
+		mysql_error "mysqld failed while attempting to check config\n\tcommand was: ${toRun[*]}\n\t$errors"
```
tianon: It looks like perhaps these `\n\t` were the reason `printf` was used directly above in `mysql_log`?  Perhaps we can use `$'...'` instead to pass literal newline and tab characters to avoid passing raw input directly to `printf`?

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+	"$@" --verbose --help --log-bin-index="$(mktemp -u)" 2>/dev/null \
+		| awk -v conf="$conf" '$1 == conf && /^[^ \t]/ { sub(/^[^ \t]+[ \t]+/, ""); print; exit }'
+	# match "datadir      /some/path with/spaces in/it here" but not "--xyz=abc\n     datadir (xyz)"
+}
+
+# Do a temporary startup of the MySQL server, for init purposes
+docker_temp_server_start() {
+	local result=0
+	%%SERVERSTARTUP%%
+	if [ "$result" != "0" ];then
```
tianon: Wouldn't something like the following be a more direct way to test the exit status of starting the temporary server?

```bash
if ! %%SERVERSTARTUP%%; then
```

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+		mysql_error "Unable to start server. Status code $result."
+	fi
+
+	# For 5.7+ the server is ready for use as soon as startup command unblocks
+	if [ "${MYSQL_MAJOR}" = "5.5" ] || [ "${MYSQL_MAJOR}" = "5.6" ]; then
```
tianon: We can drop `5.5` now, right? :innocent:

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+	fi
+}
+
+# Stop the server. When using a local socket file mysqladmin will block until
+# the shutdown is complete.
+docker_temp_server_stop() {
+	local result=0
+	mysqladmin --defaults-extra-file=<( echo "${PASSFILE}" ) shutdown -uroot --socket="${SOCKET}" || result=$?
+	if [ "$result" != "0" ]; then
+		mysql_error "Unable to shut down server. Status code $result."
```
tianon: Is the exact exit code itself relevant here?  Shouldn't there typically be some useful error output already that would be a better user clue than the exit code as to what happened anyhow?

```bash
if ! mysqladmin ...; then
```

update.sh
```diff
@@ -14,7 +14,26 @@ declare -A debianVariants=(
...
+passwordset["5.7"]="ALTER USER 'root'@'localhost' IDENTIFIED BY '\${MYSQL_ROOT_PASSWORD}';"
+passwordset["8.0"]="ALTER USER 'root'@'localhost' IDENTIFIED BY '\${MYSQL_ROOT_PASSWORD}';"
+declare -A database_init
+database_init["5.6"]='mysql_install_db --datadir="$DATADIR" --rpm --keep-my-cnf "${@:2}"'
+database_init["5.7"]='"$@" --initialize-insecure'
+database_init["8.0"]='"$@" --initialize-insecure'
+declare -A server_startup
+server_startup["5.6"]='"$@" --skip-networking --socket="${SOCKET}" \&'
+server_startup["5.7"]='"$@" --daemonize --skip-networking --socket="${SOCKET}" || result="$?"'
+server_startup["8.0"]='"$@" --daemonize --skip-networking --socket="${SOCKET}" || result="$?"'
```
tianon: Given that we've already got blocks like `if [ "${MYSQL_MAJOR}" = "5.5" ] || [ "${MYSQL_MAJOR}" = "5.6" ]; then` in the entrypoint itself, wouldn't it make sense to move these in too and thus make the script completely standalone and ready to copy instead of it being a template?

(I think the resulting code could be slightly better too, since generating Bash typically results in a bit of odd flow, which I think we want to avoid here given that folks are likely to read these scripts especially if they want to use/extend them.)

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+		mysql_ssl_rsa_setup --datadir="$DATADIR"
+		mysql_note "Certificates initialized"
+	fi
+}
+
+# Loads various settings that are used elsewhere in the script
+docker_setup_env() {
+	# Get config
+	DATADIR="$(mysql_get_config 'datadir' "$@")"
+	SOCKET="$(mysql_get_config 'socket' "$@")"
```
tianon: These are intentionally global variables, right?  Do you think it's worth doing either `export` or maybe just an explicit `declare -g DATADIR SOCKET` before setting them to make that intention more clear?

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+
+	# Write the password to the "file" the client uses
+	# the client command will use process substitution to create a file on the fly
+	# ie: --defaults-file=<( echo "${PASSFILE}" )
+	if [ ! -z "$MYSQL_ROOT_PASSWORD" ]; then
+		read -r -d '' PASSFILE <<-EOF || true
+			[client]
+			password="${MYSQL_ROOT_PASSWORD}"
+		EOF
+	fi
```
tianon: We use `echo "${PASSFILE}"` as input elsewhere but it isn't set otherwise -- do you think this might make more sense as a separate helper function that uses `MYSQL_ROOT_PASSWORD` directly to output this configuration file? (also avoiding yet another "magic" global variable?)

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+
+	# Creates a custom database and user if specified
+	if [ "$MYSQL_DATABASE" ]; then
```
tianon: I think we should probably be explicit/consistent about using `-n` or `-z` in cases like this and the test below (like we do for many above).

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+		read -r -d '' rootCreate <<-EOSQL || true
+			CREATE USER 'root'@'${MYSQL_ROOT_HOST}' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
+			GRANT ALL ON *.* TO 'root'@'${MYSQL_ROOT_HOST}' WITH GRANT OPTION ;
+		EOSQL
+	fi
+
+	docker_process_sql --database=mysql <<-EOSQL
+		-- What's done in this file shouldn't be replicated
+		--  or products like mysql-fabric won't work
+		SET @@SESSION.SQL_LOG_BIN=0;
```
tianon: Is this still relevant? (It's not used in later invocations of `docker_process_sql`.)

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+		docker_process_sql --database=mysql <<<"FLUSH PRIVILEGES ;"
+	fi
+}
+
+# Mark root user as expired so the password must be changed before anything
+# else can be done (only supported for 5.6+)
+mysql_expire_root_user() {
+	if [ ! -z "$MYSQL_ONETIME_PASSWORD" ]; then
+		if [ "${MYSQL_MAJOR}" = "5.5" ]; then
+			mysql_warn "MySQL 5.5 does not support PASSWORD EXPIRE (required for MYSQL_ONETIME_PASSWORD)"
```
tianon: More 5.5 fallbacks! :tada:

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,325 @@
...
+
+		# If container is started as root user, restart as dedicated mysql user
+		if [ "$(id -u)" = "0" ]; then
+			docker_create_db_directories
+			mysql_note "Switching to dedicated user 'mysql'"
+			exec gosu mysql "$BASH_SOURCE" "$@"
+		fi
+
+		# just in case the script was not started as root
+		docker_create_db_directories
```
tianon: If we move this up before that `if` we don't need to include it twice, right? :smile:

.template.Debian/docker-entrypoint.sh
yosifkit: I just tested the following change and our `mysql-initdb` tests worked fine (which uses `initdb.sql`). It will not work if we add the ability to run instead of source that postgres has, but then again, it wouldn't have access to our variables/functions anyway.

```diff
diff --git a/8.0/docker-entrypoint.sh b/8.0/docker-entrypoint.sh
index 0bbab71..dfea05c 100755
--- a/8.0/docker-entrypoint.sh
+++ b/8.0/docker-entrypoint.sh
@@ -51,7 +51,7 @@ docker_process_init_files() {
 	for f; do
 		case "$f" in
 			*.sh)     mysql_note "$0: running $f"; . "$f" ;;
-			*.sql)    mysql_note "$0: running $f"; docker_process_sql < "$f"; echo ;;
+			*.sql)    mysql_note "$0: running $f"; "${mysql[@]}" < "$f"; echo ;;
 			*.sql.gz) mysql_note "$0: running $f"; gunzip -c "$f" | docker_process_sql; echo ;;
 			*)        mysql_warn "$0: ignoring $f" ;;
 		esac
```

lag-linaro:
Happy Anniversary (for 5 days ago)! :champagne: 

How is this looking now?  It this still blocking #318?

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+			echo
+		fi
+	fi
+	exec "$@"
+}
+
+# This checks if the script has been sourced from elsewhere.
+# If so we don't perform any further actions
+# https://unix.stackexchange.com/a/215279
+if [ "${FUNCNAME[${#FUNCNAME[@]} - 1]}" != 'source' ]; then
```
delgod: doesn't work with centos bash
```
FUNCNAME: bad array subscript
```

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+		set -- mysqld "$@"
+	fi
+
+	# skip setup if they aren't running mysqld or want an option that stops mysqld
+	if [ "$1" = 'mysqld' ] && ! _mysql_want_help "$@"; then
+		mysql_note "Entrypoint script for MySQL Server ${MYSQL_VERSION} started."
+
+		# Load various environment variables
+		docker_setup_env "$@"
+		mysql_check_config "$@"
```
delgod: it is needed to check config before set env.
reason `docker_setup_env` function runs `mysql_get_config` which fail script if any error in config

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+		mysql_tzinfo_to_sql /usr/share/zoneinfo \
+			| sed 's/Local time zone must be set--see zic manual page/FCTY/' \
+			| MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql
+		# unset MYSQL_ROOT_PASSWORD for just docker_process_sql
+		# so that it won't try to fill in a password file when it hasn't been set yet
+	fi
+	# Generate random root password
+	if [ -n "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
+		export MYSQL_ROOT_PASSWORD="$(pwgen -1 32)"
+		mysql_note "GENERATED ROOT PASSWORD: $MYSQL_ROOT_PASSWORD"
```
delgod: old automation like ```        docker logs "$cname" | grep 'GENERATED ROOT PASSWORD' | awk '{print$4}' ```
doesn't work anymore

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+		# sed is for https://bugs.mysql.com/bug.php?id=20545
+		mysql_tzinfo_to_sql /usr/share/zoneinfo \
+			| sed 's/Local time zone must be set--see zic manual page/FCTY/' \
+			| MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql
+		# unset MYSQL_ROOT_PASSWORD for just docker_process_sql
+		# so that it won't try to fill in a password file when it hasn't been set yet
+	fi
+	# Generate random root password
+	if [ -n "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
+		export MYSQL_ROOT_PASSWORD="$(pwgen -1 32)"
```
delgod: I personally prefer `pwmake 128`, reason - it is preinstalled on centos

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+	fi
+
+	# For 5.7+ the server is ready for use as soon as startup command unblocks
+	if [ "${MYSQL_MAJOR}" = "5.6" ]; then
+		mysql_note "Waiting for server startup"
+		local i
+		for i in {30..0}; do
+			# unset MYSQL_ROOT_PASSWORD for just docker_process_sql
+			# so that it won't try to fill in a password file when it hasn't been set yet
+			if MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql <<<'SELECT 1' &> /dev/null; then
```
delgod:
```suggestion
			if MYSQL_ROOT_PASSWORD='' docker_process_sql --database=mysql <<<'SELECT 1' &> /dev/null; then
```

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+}
+
+# Initializes database with timezone info and root password, plus optional extra db/user
+docker_setup_db() {
+	# Load timezone info into database
+	if [ -z "$MYSQL_INITDB_SKIP_TZINFO" ]; then
+		# sed is for https://bugs.mysql.com/bug.php?id=20545
+		mysql_tzinfo_to_sql /usr/share/zoneinfo \
+			| sed 's/Local time zone must be set--see zic manual page/FCTY/' \
+			| MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql
```
delgod:
```suggestion
			| MYSQL_ROOT_PASSWORD='' docker_process_sql --database=mysql
```

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+		# https://unix.stackexchange.com/questions/265149/why-is-set-o-errexit-breaking-this-read-heredoc-expression/265151#265151
+		read -r -d '' rootCreate <<-EOSQL || true
+			CREATE USER 'root'@'${MYSQL_ROOT_HOST}' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
+			GRANT ALL ON *.* TO 'root'@'${MYSQL_ROOT_HOST}' WITH GRANT OPTION ;
+		EOSQL
+	fi
+
+	# unset MYSQL_ROOT_PASSWORD for just docker_process_sql
+	# so that it won't try to fill in a password file when it is just now being set
+	MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql <<-EOSQL
```
delgod:
```suggestion
	MYSQL_ROOT_PASSWORD='' docker_process_sql --database=mysql <<-EOSQL
```

.template.Debian/docker-entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+	local result=0
+	mysqladmin --defaults-extra-file=<( _mysql_passfile ) shutdown -uroot --socket="${SOCKET}" || result=$?
+	if [ "$result" != "0" ]; then
+		mysql_error "Unable to shut down server. Status code $result."
+	fi
+}
+
+# Verify that the minimally required password settings are set for new databases.
+docker_verify_minimum_env() {
+	if [ -z "$MYSQL_ROOT_PASSWORD" -a -z "$MYSQL_ALLOW_EMPTY_PASSWORD" -a -z "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
```
delgod:
```suggestion
	if [ -z "$MYSQL_ROOT_PASSWORD" ] && [ -z "$MYSQL_ALLOW_EMPTY_PASSWORD" ] && [ -z "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
```
https://github.com/koalaman/shellcheck/wiki/SC2166

5.6/docker-entrypoint.sh
```diff
@@ -40,171 +39,290 @@ file_env() {
...
+
+		# Load various environment variables
+		docker_setup_env "$@"
+		mysql_check_config "$@"
+		docker_create_db_directories
+
+		# If container is started as root user, restart as dedicated mysql user
+		if [ "$(id -u)" = "0" ]; then
+			mysql_note "Switching to dedicated user 'mysql'"
+			exec gosu mysql "$BASH_SOURCE" "$@"
```
delgod:
```suggestion
			exec gosu mysql "${BASH_SOURCE[0]}" "$@"
```

.template.Debian/docker-entrypoint.sh
yosifkit: That's confusing. While we don't have to worry about centos here, is there a Percona Server image that isn't using `centos:7`. `FUNCNAME` as an array was added in [bash 3.0](https://git.savannah.gnu.org/cgit/bash.git/tree/CHANGES?id=3ba697465bc74fab513a26dea700cc82e9f4724e#n5577) and CentOS should be plenty new enough with `4.2.46`.  :confused:

After some digging I found that `FUNCNAME` works fine within a function (like I already have in the [postgres version of this PR](https://github.com/docker-library/postgres/pull/496/files#diff-60713fb19f6c50a2473835df63ff17f9R28-R31)), so we should just use that here too.

```console
$ docker run -it --rm centos:7 bash
[root@bcfd12483c5e /]# cat test.sh 
#!/bin/bash

_is_sourced() {
	echo 'inside: ['"${FUNCNAME[@]}"']'
	echo
	[ "${FUNCNAME[${#FUNCNAME[@]} - 1]}" == 'source' ]
}

echo 'outside: ['"${FUNCNAME[@]}"']'
echo
if _is_sourced; then
	echo 'it is sourced'
fi
[root@bcfd12483c5e /]# bash test.sh
outside: []

inside: [_is_sourced main]

[root@bcfd12483c5e /]# . test.sh
outside: []

inside: [_is_sourced source]

it is sourced
```

.template.Debian/docker-entrypoint.sh
yosifkit: Yeah, I think the idea was to make the script logs look the same as `mysqld` logs, so we'll break any users with naive log grepping. It's up to you how closely you want to implement the log changes in the percona images (I don't plan to use it in the postres image).

.template.Debian/docker-entrypoint.sh
yosifkit: :man_shrugging: Using whatever is available on the base OS seems reasonable. The current MySQL images already install `pwgen` so we'll continue using what we already have.

.template.Debian/docker-entrypoint.sh
yosifkit: While we have generally moved to using `&&` over `-a`, I wanted to minimize what I added on top of @ltangvald's changes to make it easier for him to review.

We can definitely make the change in another PR.

.template.Debian/docker-entrypoint.sh
tianon: Users looking to adapt could adjust to something like `docker logs ... | grep -o 'GENERATED ROOT PASSWORD.*' | awk '{ print $4 }'`

```console
$ echo 'lol: GENERATED ROOT PASSWORD: something' | grep -o 'GENERATED ROOT PASSWORD.*' | awk '{ print $4 }'
something
```

.template.Debian/docker-entrypoint.sh
```diff
@@ -238,23 +236,31 @@ docker_setup_db() {
 	fi
 }
 
+_mysql_passfile() {
```
ltangvald: Running this every time we execute an SQL seems a bit odd, especially since it leads to the side effect of having to manipulate the user-set MYSQL_ROOT_PASSWORD variable in various places around the script.

tianon: The intent behind this is to keep the password always in-memory only instead of committing it to disk (and then also having a file we need to clean up).

ltangvald: Right, but right now it adds confusing complexity to the script, since we have to worry about the state of MYSQL_ROOT_PASSWORD in multiple places where its value should not be relevant. Maybe just add a simple flag that's set once the root password is set in the db, and have _mysql_passfile check it?

yosifkit: I see two interpretations.  We could make `_mysql_passfile` take an argument, but giving an argument to `docker_process_sql` to then pass along is a bit more complex since it currently passes everything directly to `mysql`. Since `mysql` allows a bareword of the database to connect to, any keyword we pick couldn't be used as a database name to connect to or it would have to be a flag that would never be used by `mysql`.

The other solution would be another global value, like `MYSQL_ROOT_PASSWORD_SET`, but the wrong name and users might try to set it themselves and assume root user creation would be skipped (or some other interpretation).

yosifkit: Here's my rough version using a flag. I also added `DATABASE_ALREADY_EXISTS` so that `docker_temp_server_start` could be used on an already initialized database (e.x. for an `always-initdb.d` dir).

```diff
diff --git a/.template.Debian/docker-entrypoint.sh b/.template.Debian/docker-entrypoint.sh
index 9b5ce17..85f8ec8 100755
--- a/.template.Debian/docker-entrypoint.sh
+++ b/.template.Debian/docker-entrypoint.sh
@@ -95,9 +95,13 @@ docker_temp_server_start() {
 		mysql_note "Waiting for server startup"
 		local i
 		for i in {30..0}; do
-			# unset MYSQL_ROOT_PASSWORD for just docker_process_sql
+			# only use the root password if the database has already been initializaed
 			# so that it won't try to fill in a password file when it hasn't been set yet
-			if MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql <<<'SELECT 1' &> /dev/null; then
+			extraArgs=()
+			if [ -z "$DATABASE_ALREADY_EXISTS" ]; then
+				extraArgs+=( '--dont-use-mysql-root-password' )
+			fi
+			if docker_process_sql "${extraArgs[@]}" --database=mysql <<<'SELECT 1' &> /dev/null; then
 				break
 			fi
 			sleep 1
@@ -168,19 +172,29 @@ docker_setup_env() {
 	file_env 'MYSQL_USER'
 	file_env 'MYSQL_PASSWORD'
 	file_env 'MYSQL_ROOT_PASSWORD'
+
+	declare -g DATABASE_ALREADY_EXISTS
+	if [ -d "$DATADIR/mysql" ]; then
+		DATABASE_ALREADY_EXISTS='true'
+	fi
 }
 
 # Execute sql script, passed via stdin
-# usage: docker_process_sql [mysql-cli-args]
+# usage: docker_process_sql [--dont-use-mysql-root-password] [mysql-cli-args]
 #    ie: docker_process_sql --database=mydb <<<'INSERT ...'
-#    ie: docker_process_sql --database=mydb <my-file.sql
+#    ie: docker_process_sql --dont-use-mysql-root-password --database=mydb <my-file.sql
 docker_process_sql() {
+	passfileArgs=()
+	if [ '--dont-use-mysql-root-password' = "$1" ]; then
+		passfileArgs+=( "$1" )
+		shift
+	fi
 	# args sent in can override this db, since they will be later in the command
 	if [ -n "$MYSQL_DATABASE" ]; then
 		set -- --database="$MYSQL_DATABASE" "$@"
 	fi
 
-	mysql --defaults-file=<( _mysql_passfile ) --protocol=socket -uroot -hlocalhost --socket="${SOCKET}" "$@"
+	mysql --defaults-file=<( _mysql_passfile "${passfileArgs[@]}") --protocol=socket -uroot -hlocalhost --socket="${SOCKET}" "$@"
 }
 
 # Initializes database with timezone info and root password, plus optional extra db/user
@@ -190,9 +204,8 @@ docker_setup_db() {
 		# sed is for https://bugs.mysql.com/bug.php?id=20545
 		mysql_tzinfo_to_sql /usr/share/zoneinfo \
 			| sed 's/Local time zone must be set--see zic manual page/FCTY/' \
-			| MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql
-		# unset MYSQL_ROOT_PASSWORD for just docker_process_sql
-		# so that it won't try to fill in a password file when it hasn't been set yet
+			| docker_process_sql --dont-use-mysql-root-password --database=mysql
+			# tell docker_process_sql to not use MYSQL_ROOT_PASSWORD since it is not set yet
 	fi
 	# Generate random root password
 	if [ -n "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
@@ -211,9 +224,8 @@ docker_setup_db() {
 		EOSQL
 	fi
 
-	# unset MYSQL_ROOT_PASSWORD for just docker_process_sql
-	# so that it won't try to fill in a password file when it is just now being set
-	MYSQL_ROOT_PASSWORD= docker_process_sql --database=mysql <<-EOSQL
+	# tell docker_process_sql to not use MYSQL_ROOT_PASSWORD since it is just now being set
+	docker_process_sql --dont-use-mysql-root-password --database=mysql <<-EOSQL
 		-- What's done in this file shouldn't be replicated
 		--  or products like mysql-fabric won't work
 		SET @@SESSION.SQL_LOG_BIN=0;
@@ -248,7 +260,7 @@ _mysql_passfile() {
 	# echo the password to the "file" the client uses
 	# the client command will use process substitution to create a file on the fly
 	# ie: --defaults-file=<( _mysql_passfile )
-	if [ -n "$MYSQL_ROOT_PASSWORD" ]; then
+	if [ '--dont-use-mysql-root-password' != "$1" ] && [ -n "$MYSQL_ROOT_PASSWORD" ]; then
 		cat <<-EOF
 			[client]
 			password="${MYSQL_ROOT_PASSWORD}"
@@ -301,8 +313,8 @@ _main() {
 			exec gosu mysql "$BASH_SOURCE" "$@"
 		fi
 
-		# If this is true then there's no database, and it needs to be initialized
-		if [ ! -d "$DATADIR/mysql" ]; then
+		# there's no database, so it needs to be initialized
+		if [ -z "$DATABASE_ALREADY_EXISTS" ]; then
 			docker_verify_minimum_env
 			docker_init_database_dir "$@"
```

ltangvald: Yeah, I think this could work (I still have some niggling doubts, but don't see any obvious better ways to handle it, and would rather not have a 2nd anniversary for this thing).

lag-linaro:
Surely, surely, this is ready to go now.

Last, little, push ...

.template.Debian/docker-entrypoint.sh
yosifkit: > Is this still relevant? (It's not used in later invocations of docker_process_sql.)

I think this is the only outstanding question and since it is about code that isn't new, I think we can put it in its own issue.

tianon:
Agreed, let's take this across the finish line and dial in anything else in follow-ups. :+1:

(I think we've discussed/evaluated the high level bits enough that we're relatively safe.)

lag-linaro:
:clap:

PR #473, opened 2018-08-12
Added MYSQL_UID and MYSQL_GID optional arguments / env. variables

optimistiCli:
Passing uid and gid to “docker build” allows for consistent mysql user / group across host and image. 

Also some hosts might have different (compared to debian) ideas about what uid / gid ranges are reserved for system users.

Hope this helps :-)

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -1,11 +1,10 @@
 FROM debian:stretch-slim
 
+# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 ARG MYSQL_UID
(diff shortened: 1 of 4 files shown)


optimistiCli:
Travis CI keeps failing on network timeouts. I guess it's time for me to give up trying to get through this hurdle :-)

yosifkit:
This is unnecessary. Since #161 the image can be run as whatever user id is desired via `--user` and `user:` to `docker run` and in a `docker compose` file, respectively.

The only assumption is that if the container is given a local folder (`-v /some/host/path:/var/lib/mysql`), then the folder would need to have appropriate permissions for the user specified in the docker run or compose file.

PR #500, opened 2018-10-04
escape database underscore wildcard

Istador:
if the $MYSQL_DATABASE environment variable contains underscores ("_")
they are now escaped, so that the GRANT matches exactly the given
database name.

### DIFF
--- a/.template.Debian/docker-entrypoint.sh
+++ b/.template.Debian/docker-entrypoint.sh
@@ -284,7 +284,7 @@ docker_setup_db() {
 
 		if [ -n "$MYSQL_DATABASE" ]; then
 			mysql_note "Giving user ${MYSQL_USER} access to schema ${MYSQL_DATABASE}"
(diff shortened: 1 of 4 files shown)


Istador:
I force pushed a new version, that is no longer in conflict with the changes made to `master` since I created this PR.

tianon:
Sorry for the huge delay :grimacing: 

I've updated this to include a change to the templated version of this same script.

For posterity, here's the relevant MySQL documentation: (https://dev.mysql.com/doc/refman/8.0/en/grant.html#grant-quoting)

> The `_` and `%` wildcards are permitted when specifying database names in `GRANT` statements that grant privileges at the database level (`GRANT ... ON db_name.*`). This means, for example, that to use a `_` character as part of a database name, specify it as `\_` in the `GRANT` statement, to prevent the user from being able to access additional databases matching the wildcard pattern (for example, `GRANT ... ON `foo\_bar`.* TO ...`).

PR #506, opened 2018-10-18
Adding progress report functions to the script and db imports

JamesonNetworks:


JamesonNetworks:
Addresses Issue #507

yosifkit:
Thanks for the contribution!

I'd rather not add this sort of complexity to the image when users can do this in their own scripts if they need an indication of progress on a large initialization.

PR #517, opened 2018-11-22
fix issue of 'gpg: keyserver receive failed: Cannot assign requested …

DrSnowbird:
fix issue of 'gpg: keyserver receive failed: Cannot assign requested address


tianon:
See https://github.com/docker-library/mysql/pull/459 for the way we fix this issue already.

PR #536, opened 2019-02-02
Touch $DATADIR/docker-entrypoint-complete after loading complete

jamielennox:
When the mysql container is autoloading data from
/docker-entrypoint-initdb.d/ it's difficult to know when the load has
completed. This is particularly true in a kubernetes
liveliness/readiness check where we don't know if the container is ready
to serve yet.

After the load has finished we touch a file that we can test for the
presence of before responding the container is ready.

### DIFF
--- a/5.5/docker-entrypoint.sh
+++ b/5.5/docker-entrypoint.sh
@@ -205,6 +205,7 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 		echo 'MySQL init process done. Ready for start up.'
 		echo
 	fi
+	touch "$DATADIR/docker-entrypoint-complete"
 fi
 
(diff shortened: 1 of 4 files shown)


jamielennox:
The downside of putting it in $DATADIR is that it will end up in the volume if one is mounted, I was going to just keep it in / to be per container. Unfortunately entrypoint.sh runs as the mysql user and the $DATADIR is the only place I think we have write access to. (/tmp seems wrong).

yosifkit:
This is not needed. The entrypoint is designed so that if it is available via tcp, then it is done loading `docker-entrypoint-initdb.d` files. If your readiness checks via IP across the network (or a local script uses `127.0.0.1` to check for connection), then it will just work.

PR #548, opened 2019-03-10
Fix that creating 'root'@'${MYSQL_ROOT_HOST}' will be failed when MYSQL_ROOT_HOST isn't '%'

flytreeleft:
`FLUSH PRIVILEGES` needs to be executed after creating or updating any user.


flytreeleft:
Close it, because the commit comment loses some information, so I should create a new PR for completing it.

flytreeleft:
Go to the new PR #549

PR #553, opened 2019-03-22
bugfix for entrypoint path.

crazy-canux:
entrypoint not working because the typo path.

### DIFF
--- a/8.0/Dockerfile
+++ b/8.0/Dockerfile
@@ -68,7 +68,7 @@ VOLUME /var/lib/mysql
 # Config files
 COPY config/ /etc/mysql/
 COPY docker-entrypoint.sh /usr/local/bin/
-RUN ln -s usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
+RUN ln -s /usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
 ENTRYPOINT ["docker-entrypoint.sh"]
 
 EXPOSE 3306 33060


tianon:
That's intentional; it's a relative symlink, not a typo.

crazy-canux:
Yes. It's not safe.
In my case the original image not working.
But use absolute path works.

tianon:
This symlink only exists for backwards compatibility -- in normal image operation, it's completely unused.  Also, there's nothing inherently "more safe" about an absolute vs a relative symlink, so you might need to elaborate more there.

Additionally, I cannot reproduce any error while using the compatibility symlink on the existing image:

```console
$ docker pull mysql:8
8: Pulling from library/mysql
Digest: sha256:4589ba2850b93d103e60011fe528fc56230516c1efb4d3494c33ff499505356f
Status: Image is up to date for mysql:8

$ docker run -it --rm -e MYSQL_RANDOM_ROOT_PASSWORD=1 mysql:8 /entrypoint.sh mysqld
Initializing database
2019-03-22T20:16:44.992637Z 0 [Warning] [MY-011070] [Server] 'Disabling symbolic links using --skip-symbolic-links (or equivalent) is the default. Consider not using this option as it' is deprecated and will be removed in a future release.
2019-03-22T20:16:44.992694Z 0 [System] [MY-013169] [Server] /usr/sbin/mysqld (mysqld 8.0.15) initializing of server in progress as process 32
2019-03-22T20:16:46.551402Z 5 [Warning] [MY-010453] [Server] root@localhost is created with an empty password ! Please consider switching off the --initialize-insecure option.
2019-03-22T20:16:47.259256Z 0 [System] [MY-013170] [Server] /usr/sbin/mysqld (mysqld 8.0.15) initializing of server has completed
Database initialized
MySQL init process in progress...
...
```

tianon:
Closing given that we cannot reproduce an issue (as shown above).

PR #578, opened 2019-07-13
Enable Toggling of Random Root Password

zanbaldwin:
**Enable a random root password to not be generated even when the `MYSQL_RANDOM_ROOT_PASSWORD` environment variable is set by setting it to a value of `no`.**

According to the Docker image documentation, a random root password can be generated:

> ### `MYSQL_RANDOM_ROOT_PASSWORD`
> 
> This is an optional variable. Set to `yes` to generate a random initial password for the root user (using `pwgen`). The generated root password will be printed to stdout (`GENERATED ROOT PASSWORD: .....`).

However a random root password is generated whenever that environment variable is present, regardless of it's value. There may be times when this variable is present but it's not desired to have a root password generated randomly.

# Example Use Case

```yaml
# docker-compose.yaml (Production)
services:
    mysql:
        image: 'mysql:8.0.15'
        environment:
            MYSQL_ALLOW_EMPTY_PASSWORD: 'this value isnt checked'
            MYSQL_RANDOM_ROOT_PASSWORD: 'yes'
```

```yaml
# docker-compose.override.yaml (Development)
services:
    mysql:
        environment:
            MYSQL_RANDOM_ROOT_PASSWORD: 'no'
```

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -133,7 +133,7 @@ if [ "$1" = 'mysqld' -a -z "$wantHelp" ]; then
 			mysql_tzinfo_to_sql /usr/share/zoneinfo | sed 's/Local time zone must be set--see zic manual page/FCTY/' | "${mysql[@]}" mysql
 		fi
 
-		if [ ! -z "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
+		if [ ! -z "$MYSQL_RANDOM_ROOT_PASSWORD" -a "$MYSQL_RANDOM_ROOT_PASSWORD" != 'no' ]; then
 			export MYSQL_ROOT_PASSWORD="$(pwgen -1 32)"
 			echo "GENERATED ROOT PASSWORD: $MYSQL_ROOT_PASSWORD"
 		fi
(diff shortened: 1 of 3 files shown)


yosifkit:
Apologies that the documentation is not clear. The code checks that the variable is not empty (which is our intended behavior). Fortunately the override.yml can make it the empty string again (or appear undefined, which is equivalent in the `! -z` test):

```yml
# docker-compose.yml
version: '3.3'

services:
    bash:
        image: bash:5.0
        environment:
            - FOO=original
            - BAR=original
        command: [ "sh", "-c", "env | grep -iE 'foo|bar'" ]

## override.yml
version: '3.3'

services:
    bash:
        environment:
            FOO: ''
# and this line makes it not even appear via env
            BAR:
```

```console
$ # without override
$ docker-compose run bash
BAR=original
FOO=original
$ # with override
$ docker-compose run bash
FOO=
$ # and inspecting the container
$ docker container inspect xyz_container
...
            "Env": [
                "FOO=",
                "BAR",
...
```

zanbaldwin:
Amazing, where can I make a pull request against the docs displayed on [Docker Hub](https://hub.docker.com/_/mysql)?

tianon:
> -	**Source of this description**:  
> 	[docs repo's `mysql/` directory](https://github.com/docker-library/docs/tree/master/mysql) ([history](https://github.com/docker-library/docs/commits/master/mysql))

PR #580, opened 2019-07-28
Use https in update.sh

J0WI:


update.sh
```diff
@@ -38,7 +38,7 @@ for version in "${versions[@]}"; do
 			-e 's/^(ENV MYSQL_VERSION) .*/\1 '"$fullVersion"'/' \
 			-e 's/^(ENV MYSQL_MAJOR) .*/\1 '"$version"'/' \
 			-e 's/^(FROM) .*/\1 debian:'"$debianVariant"'/' \
-			-e 's!(http://repo.mysql.com/apt/debian/) [^ ]+!\1 '"$debianSuite"'!' \
+			-e 's!(https://repo.mysql.com/apt/debian/) [^ ]+!\1 '"$debianSuite"'!' \
```
tianon: This won't match anything in the `Dockerfile` :confused:

J0WI: That was too much 😄

Issue #589, opened 2019-09-10
Show progress while importing SQL data


mojoaxel:
_Feature Request:_

It would be nice to show a progress bar while importing e.g. `*.sql` data.

For that the `process_init_file` function in the `docker-entrypoint.sh` needs to be adopted.

yosifkit:
Closing since related PR was closed.

PR #590, opened 2019-09-10
show progress while importing SQL data

mojoaxel:
implements #589

_I have to admit that the changes are untested but I expect it to work!_


yosifkit:
We'd rather not install an extra unnecessary utility in the image (even if it is rather small).  Users are free customize their scripts to give themselves progress for large imports (or even do them outside `initdb.d` scripts entirely).

PR #601, opened 2019-10-16
Revert to not using daemonize flag for temporary startup of 5.7 server

ltangvald:
Daemonize lets us avoid a manual wait loop for the server to be ready, but
for 5.7 it causes a log redirection error when the image is started with -t,
so for now just revert 5.7 image to using the wait loop.


tianon:
LGTM, but we do need to run `./update.sh` and commit the copied result so that Travis tests the right thing :sweat_smile:

PR #605, opened 2019-10-30
Fix config update

Enapiuz:
At least on macOS you have to specify `-e` option for echo, otherwise you won't get line separators.


tianon:
This is actually working fine as-is due to a quirk of `/bin/sh` in Debian:

```console
$ docker pull mysql:5.7
5.7: Pulling from library/mysql
Digest: sha256:44b33224e3c406bf50b5a2ee4286ed0d7f2c5aec1f7fdb70291f7f7c570284dd
Status: Image is up to date for mysql:5.7

$ docker run --rm mysql:5.7 cat /etc/mysql/conf.d/docker.cnf
[mysqld]
skip-host-cache
skip-name-resolve
```

Issue #571, opened 2019-06-15
Root password ending on '\J' causes a crash


ArthurHNL:
When creating an instance of the `mysql:5.7` container, with the following variables from an env file (using docker compose):

```
root@demo01:/conf/docker# cat envfile.env | grep 'MYSQL'
MYSQL_DATABASE=somevalue
MYSQL_USER=sqluser
MYSQL_PASSWORD=(6T+B\`2@`Bf]cqF
MYSQL_ROOT_PASSWORD=;*{q#'Br9+qxh2\J
MYSQL_ROOT_HOST=%
```

MySQL crashes with the following output in the log:

```
Initializing database
2019-06-15T15:23:59.415039Z 0 [Warning] TIMESTAMP with implicit DEFAULT value is deprecated. Please use --explicit_defaults_for_timestamp server option (see documentation for more details).
2019-06-15T15:24:00.127942Z 0 [Warning] InnoDB: New log files created, LSN=45790
2019-06-15T15:24:00.250132Z 0 [Warning] InnoDB: Creating foreign key constraint system tables.
2019-06-15T15:24:00.310568Z 0 [Warning] No existing UUID has been found, so we assume that this is the first time that this server has been started. Generating a new UUID: 99a9caa4-8f81-11e9-acf7-0242ac140002.
2019-06-15T15:24:00.311781Z 0 [Warning] Gtid table is not ready to be used. Table 'mysql.gtid_executed' cannot be opened.
2019-06-15T15:24:00.312706Z 1 [Warning] root@localhost is created with an empty password ! Please consider switching off the --initialize-insecure option.
Database initialized
Initializing certificates
Generating a RSA private key
.................................................................+++++
.................................................................................+++++
unable to write 'random state'
writing new private key to 'ca-key.pem'
-----
Generating a RSA private key
...........+++++
..........................................+++++
unable to write 'random state'
writing new private key to 'server-key.pem'
-----
Generating a RSA private key
.........................................................................+++++
................+++++
unable to write 'random state'
writing new private key to 'client-key.pem'
-----
Certificates initialized
MySQL init process in progress...
2019-06-15T15:24:06.640728Z 0 [Warning] TIMESTAMP with implicit DEFAULT value is deprecated. Please use --explicit_defaults_for_timestamp server option (see documentation for more details).
2019-06-15T15:24:06.648939Z 0 [Note] mysqld (mysqld 5.7.26) starting as process 89 ...
[...]

Issue #338, opened 2017-10-26
Creating Multiple Databases without init script


scrummyin:
I found a bug/feature. I was able to create multiple databases without the need for my own dockerfile. I kinda abused the docker-entrypoint.sh. Below is a terminal session that shows how I did it

```
$ cat elist
MYSQL_DATABASE=one`; CREATE DATABASE IF NOT EXISTS two; CREATE DATABASE IF NOT EXISTS `three
MYSQL_ROOT_PASSWORD=pass
$ docker run -p 127.0.0.1:3306:3306 --env-file elist mysql
```

I'm not sure want to do with this information, but it works as a quick hack to get multiple databases.

tianon:
Heh, this is a cute hack -- I'm not sure what to do with it either.  Either leave it as an easter egg or try to do some form of `mysql_real_escape_string` in Bash? (yuck)

I'm inclined to simply leave it as an undocumented thing that happens to work as a consequence of how our script works and leave it alone (since we'd likely have to add more dependencies and thus make the image bigger to attempt to "fix" it properly and having backtick in a database name should be pretty rare and those folks should have to think long and hard about what the heck they're doing).

tianon:
Since this isn't something we've deemed worth fixing, I'm going to close -- the issue will hopefully serve as a good enough reference for folks looking to take advantage of this accidental "feature" (which is really a bug, to be explicitly clear). :+1:

seekcx:
- [x] Multiple database support.

Awesome!
This is exactly what I want.

dmitrym0:
This is cute, but surely there's a cleaner way to create multiple database on init? What about permission grants?

tianon:
The cleaner way is to use a proper SQL file (or shell script) in `/docker-entrypoint-initdb.d` that performs the steps necessary to set up the database in the way that you need.

xandriaw:
@tianon can I just have two .sql scripts, each with 'create database' commands in them or is there something more necessary?

tianon:
That should be plenty, yep!

glensc:
https://xkcd.com/327/ :D

PR #613, opened 2019-12-08
strip newline from all passed variables

OndrejHome:
This prevents weird errors when attributes get shifted to newline and are executed by docker-entrypoint.sh. This PR removes the newline and when newline is stripped a 'Warning' is printed.

Short version: `docker-entrypoint.sh` scripts breaks (as seen later in examples of this PR) when 'newline' is present with rather ugly error that is hard to analyze without checking the code of script.
Better error message or automated removal of 'newline' would be more desirable. This PR proposes automated 'newline' removal.

Long version:
I have run into this when trying to get image working with kubernetes and erroneously generated 'kubernetes secret' that contained the newline at the end. From looking at the logs the error that I have made required me to do debugging of image as there was no good message on what is going on. I believe there is no practical use case for having 'newline' in any of variables (if this is not true please correct me). 

After the change proposed here the mysql image would progress with only printing warning when 'newline' is included which would make this work in most cases. With kubernetes the newline is preserved at the end of string which makes this invisible to user if such mistake is done. 
Alternative approach to this would be to "hard fail" with error that 'newline' was encountered in some variable as that will most probably never be desired and user should be aware of it. 

Main idea for creating this PR is to avoid unnecessary image debugging when 'newline' is accidentally included in any of the variables. 

Please let me know what you think about this change and also let me know there are any changes needed for this to be merged. Thank you!

How to reproduce issue:
~~~
# a="aa
bb"
# docker run --name t56 -e MYSQL_ROOT_PASSWORD="$a" -d mysql:5.6 --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
xxxxxxxx
# docker logs -f xxxxxxxx
~~~

**mysql:5.6**
~~~
...
2019-12-08 08:18:54+00:00 [Note] [Entrypoint]: Temporary server started.
Warning: Unable to load '/usr/share/zoneinfo/iso3166.tab' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/leap-seconds.list' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/zone.tab' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/zone1970.tab' as time zone. Skipping it.
2019-12-08 08:18:56 96 [Warning] 'proxies_priv' entry '@ root@0b2ef1dce921' ignored in --skip-name-resolve mode.

2019-12-08 08:18:56+00:00 [Note] [Entrypoint]: Stopping temporary server
mysqladmin: unknown option '--bb"'
2019-12-08 08:18:56+00:00 [ERROR] [Entrypoint]: Unable to shut down server.
~~~

**mysql:5.7**
~~~
...
2019-12-08 08:20:09+00:00 [Note] [Entrypoint]: Temporary server started.
Warning: Unable to load '/usr/share/zoneinfo/iso3166.tab' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/leap-seconds.list' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/zone.tab' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/zone1970.tab' as time zone. Skipping it.

2019-12-08 08:20:17+00:00 [Note] [Entrypoint]: Stopping temporary server
mysqladmin: [ERROR] unknown option '--bb"'
2019-12-08 08:20:17+00:00 [ERROR] [Entrypoint]: Unable to shut down server.
~~~

**mysql:8.0**
~~~
...
2019-12-08 08:22:48+00:00 [Note] [Entrypoint]: Temporary server started.
2019-12-08T08:22:48.929209Z 0 [System] [MY-011323] [Server] X Plugin ready for connections. Socket: '/var/run/mysqld/mysqlx.sock'
Warning: Unable to load '/usr/share/zoneinfo/iso3166.tab' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/leap-seconds.list' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/zone.tab' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/zone1970.tab' as time zone. Skipping it.

2019-12-08 08:22:54+00:00 [Note] [Entrypoint]: Stopping temporary server
mysqladmin: [ERROR] unknown option '--bb"'.
2019-12-08 08:22:54+00:00 [ERROR] [Entrypoint]: Unable to shut down server.
~~~

After fix the mysql can start properly and following can be seen in logs if there was any newline stripped (including the name of variable from which it was stripped so user can fix it):
~~~
2019-12-08 08:38:22+00:00 [Warn] [Entrypoint]: Stripping newline from MYSQL_ROOT_PASSWORD.
~~~

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -35,6 +35,12 @@ file_env() {
 	elif [ "${!fileVar:-}" ]; then
 		val="$(< "${!fileVar}")"
 	fi
+        # remove newline from values which would cause troubles later if present
+        new_val=$(echo "$val"|tr -d '\n')
+        if [ "$new_val" != "$val" ]; then
+                mysql_warn "Stripping newline from ${var}."
+                val="$new_val"
+        fi
 	export "$var"="$val"
 	unset "$fileVar"
 }
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -35,6 +35,12 @@ file_env() {
 	elif [ "${!fileVar:-}" ]; then
 		val="$(< "${!fileVar}")"
 	fi
+        # remove newline from values which would cause troubles later if present
+        new_val=$(echo "$val"|tr -d '\n')
+        if [ "$new_val" != "$val" ]; then
+                mysql_warn "Stripping newline from ${var}."
+                val="$new_val"
+        fi
 	export "$var"="$val"
 	unset "$fileVar"
 }
--- a/8.0/docker-entrypoint.sh
+++ b/8.0/docker-entrypoint.sh
@@ -35,6 +35,12 @@ file_env() {
 	elif [ "${!fileVar:-}" ]; then
 		val="$(< "${!fileVar}")"
 	fi
+        # remove newline from values which would cause troubles later if present
+        new_val=$(echo "$val"|tr -d '\n')
+        if [ "$new_val" != "$val" ]; then
+                mysql_warn "Stripping newline from ${var}."
+                val="$new_val"
+        fi
 	export "$var"="$val"
 	unset "$fileVar"
 }


yosifkit:
Seems to be related to these issues: https://github.com/docker-library/mariadb/issues/183 and https://github.com/docker-library/mysql/issues/571

While this is one possible way to handle "bad" characters and is targeted at fixing the "Option File" created by `docker_process_sql` (and `_mysql_passfile`), I'd rather not "silently" change what the user passed in. Many won't notice the log message and then wonder why they cannot connect.

I see two main alternatives to address escaping (passwords specifically*): fail on bad values or escape everything. Escaping everything will probably lead us to things like implementing `mysql_real_escape_string` in bash (and we'd rather not maintain that). On the other hand, early failing on any input that just needs an escape hurts user experience.  I think we can meet somewhere in the middle that will allow most character sequences, while failing on more complex ones without maintaining a full `mysql_real_escape_string` in bash.

:arrow_up: WDYT @tianon and @ltangvald?  Maybe something like [wordpress](https://github.com/docker-library/wordpress/blob/532bb60b1b789534a2f745f8de01afb8f6256c18/docker-entrypoint.sh#L181-L186) uses when escaping things for `sed`?

-------------------------------------------

Side note: https://github.com/docker-library/mariadb/issues/183 discussed `printf %q`, but that was only in the context of SQL escaping and may not be enough since the password in the "option file" is interpreted differently than in SQL (https://dev.mysql.com/doc/refman/8.0/en/option-files.html#option-file-syntax).

\* I say passwords specifically since most other environment variables like user or database are unlikely to contain characters that need escaping (and have accidental "features" https://github.com/docker-library/mysql/issues/338).

OndrejHome:
Hi @yosifkit, Thanks for feedback. 

> I see two main alternatives to address escaping (passwords specifically*): fail on bad values or escape everything.

For simplicity I would suggest 'failing on bad values' and letting the user know what bad values are. If the container fails to start and last message in its logs looks like `[ERROR] [Entrypoint]: variable MYSQL_ROOT_PASSWORD contains restricted character(s) like '$/\J\n...'` then I would expect user to either "brute-force" what they can/cannot put into variable or complain that their favourite special thing is not allowed (which can lead into issue that would eventually add character if considered viable). In both of these cases it would lead to improvement in user experience as more things will get allowed. For backward-compatibility this can be done for example only for one type of variables (either ones passed via 'env' or ones from 'file').
If interested let me know and I can draft code example for above idea.

ltangvald:
I was going to suggest printf %q (we used it in e.g. the upstream debian packages to escape input during the installation), but if you give it a multiline input it wraps it in $'', which causes a syntax error in SQL

ltangvald:
I like the idea of doing input validation, i.e. fail as it does now, but with a more helpful error message.

tianon:
That sounds like fun -- is there somewhere we can reference for what the "canonically supported" set of characters / restrictions are? :smile:

ltangvald:
There is a list of special characters here https://dev.mysql.com/doc/refman/8.0/en/string-literals.html
But detecting them reliably might be a bit tricky. We could maybe use printf %q to escape the "normal" ones, and try to just detect tricker cases like newlines?

PR #616, opened 2019-12-18
adding a walk_dir method to loop over sql script inside sub folders

hamzawey:
Added a walk_dir method to loop over sql scripts inside sub folders such that 
- 001/
    - script001_1.sql
    - script001_2.sql
- 002/
    - script002_1.sql
    - script002_2.sql


yosifkit:
Duplicate of #192 #193.

Unfortunately this is not something we want to add at this time.

Related issues https://github.com/docker-library/postgres/issues/179, https://github.com/docker-library/postgres/issues/605

PR #627, opened 2020-01-18
update debian base version to buster

inductor:
It's been a while since buster released

### DIFF
--- a/5.6/Dockerfile
+++ b/5.6/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch-slim
+FROM debian:buster-slim
 
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN groupadd -r mysql && useradd -r -g mysql mysql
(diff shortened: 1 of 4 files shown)


yosifkit:
> E: Version '5.6.47-1debian10' for 'mysql-server' was not found

It seems that 5.6 isn't packaged for Buster, so it will have to stay on Stretch. I think just adjusting line 14 in `update.sh` to be `[5.6]='stretch'` and running `update.sh` should do it.

@tianon and @ltangvald, any other comments or concerns on bumping the OS?

tianon:
Do we need to worry about anything like https://github.com/docker-library/postgres/issues/582? (locale changes in glibc that manifest in buster+)

inductor:
@yosifkit @tianon

Updated the shell!

tianon:
> Do we need to worry about anything like [docker-library/postgres#582](https://github.com/docker-library/postgres/issues/582)? (locale changes in glibc that manifest in buster+)

friendly ping @ltangvald -- is this something we need to worry about, or does MySQL handle locales differently such that it's not a concern?

ltangvald:
Hm, I vaguely remember thinking there might be some issue with updating from stretch to buster, but I can't remember what it might have been (maybe just that we dropped 32bit support for buster, which wouldn't matter)

ltangvald:
Checked with the devs, and no, MySQL should not be affected by that PG issue.

tianon:
That's excellent! 🥳

So we're good to go, or you want to rack your brain a bit to try and remember what you had in mind? 😅

ltangvald:
No, I think we're good to go :)

PR #643, opened 2020-02-27
Restore fix for silently skipped init scripts

tianon:
This was fixed previously in https://github.com/docker-library/mysql/pull/453 but lost in https://github.com/docker-library/mysql/pull/471. :sweat_smile:


yosifkit:
I moved the check earlier to help prevent an initialized database that failed to apply all `initdb.d` scripts.

PR #208, opened 2016-09-03
Allow override of DATADIR for database init

Ulexus:
This represents what I view as a minimally-invasive solution to #207 

A more "complete" solution would add a `--datadir` argument to the execution if there is not already one AND `DATADIR` is set to something other than the default, but this modification simply makes it _possible_ to use this container, unmodified, with a non-default data directory.

Examples:

Preserve the status quo:

```
~/c/p/m/5.7 (allow-datadir-overrride) $ docker run --rm --name=statusquo -e MYSQL_ROOT_PASSWORD=testme mysql:test
Initializing database
Database initialized
MySQL init process in progress...
Warning: Unable to load '/usr/share/zoneinfo/Factory' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/iso3166.tab' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/leap-seconds.list' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/posix/Factory' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/right/Factory' as time zone. Skipping it.
Warning: Unable to load '/usr/share/zoneinfo/zone.tab' as time zone. Skipping it.


MySQL init process done. Ready for start up.

```

Check:

```
~/c/p/m/5.7 (allow-datadir-overrride) $ docker exec -it statusquo ls -l /var/lib/mysql
total 188424
-rw-r----- 1 mysql mysql       56 Sep  3 17:56 auto.cnf
-rw-r----- 1 mysql mysql     1329 Sep  3 17:56 ib_buffer_pool
-rw-r----- 1 mysql mysql 50331648 Sep  3 17:56 ib_logfile0
-rw-r----- 1 mysql mysql 50331648 Sep  3 17:56 ib_logfile1
-rw-r----- 1 mysql mysql 79691776 Sep  3 17:56 ibdata1
-rw-r----- 1 mysql mysql 12582912 Sep  3 17:56 ibtmp1
drwxr-x--- 1 mysql mysql     2298 Sep  3 17:56 mysql
drwxr-x--- 1 mysql mysql     5638 Sep  3 17:56 performance_schema
drwxr-x--- 1 mysql mysql     6656 Sep  3 17:56 sys
```

Non-default data directory:

```
[...]

PR #660, opened 2020-04-16
Allow to configure DATADIR via env variable

ostrolucky:
This is useful to do when mounting volumes. In some cases, e.g. with named volumes, it's not possible to mount sub-folder<sup>[1](https://github.com/moby/moby/issues/32582)</sup>. 

This limitation prevents reuse of same volume. For example, If both mongodb and mysql are configured to use shared volume as its DB folder, mysql will refuse to start, since its target DB folder is not empty (`[ERROR] --initialize specified but the data directory has files in it. Aborting.`). This patch allows to solve this by mounting shared volume to different directory than `/var/lib/mysql` and define `$MYSQL_DATADIR` to subfolder of this shared volume.

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -136,11 +136,11 @@ docker_create_db_directories() {
 
 	# TODO other directories that are used by default? like /var/lib/mysql-files
 	# see https://github.com/docker-library/mysql/issues/562
-	mkdir -p "$DATADIR"
+	mkdir -p "$MYSQL_DATADIR"
 
 	if [ "$user" = "0" ]; then
 		# this will cause less disk access than `chown -R`
-		find "$DATADIR" \! -user mysql -exec chown mysql '{}' +
+		find "$MYSQL_DATADIR" \! -user mysql -exec chown mysql '{}' +
 	fi
 }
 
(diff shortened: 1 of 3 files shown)


tianon:
Duplicate of #208

ostrolucky:
Indeed. Although this is more elegant since it doesn't increase complexity and no fiddling with mysql config files is necessary.

tianon:
I'm not sure I follow?  This can be set via the command-line as well (and indeed, that's how I set it for instances where I need this):

```console
$ docker pull mysql:5.7
5.7: Pulling from library/mysql
Digest: sha256:7924ec534b844f9c7c0547fa95265516e8adf19ca475cffe911202037392d264
Status: Image is up to date for mysql:5.7
docker.io/library/mysql:5.7

$ docker run -d -e MYSQL_RANDOM_ROOT_PASSWORD=1 --name test mysql:5.7 --datadir /some/random/directory/location
26b9f3fbcc20b64d453f7b3230a37cdb06a1d4c06cc516d4dcc4b7be334800bb

$ docker logs --tail=1 test
Version: '5.7.29'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306  MySQL Community Server (GPL)

$ docker diff test
C /run
C /run/mysqld
A /run/mysqld/mysqld.pid
A /run/mysqld/mysqld.sock
A /run/mysqld/mysqld.sock.lock
A /some
A /some/random
A /some/random/directory
A /some/random/directory/location
A /some/random/directory/location/client-key.pem
A /some/random/directory/location/performance_schema
A /some/random/directory/location/performance_schema/events_stages_history_long.frm
A /some/random/directory/location/performance_schema/events_waits_summary_by_user_by_event_name.frm
A /some/random/directory/location/performance_schema/cond_instances.frm
...
```

PR #669, opened 2020-05-03
Dump on container termination

ruslanzelinskyy:
I think there is a problem with saving db after container stopped.
I have no possibility to change that behaviour with docker-compose in the new dockerfile
So i suggest to add some trap signals to save db data before container stops

### DIFF
--- a/8.0/docker-entrypoint.sh
+++ b/8.0/docker-entrypoint.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -eo pipefail
 shopt -s nullglob
-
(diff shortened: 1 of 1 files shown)


ruslanzelinskyy:
Sorry missed to change root pass and urser to env vars
Also i suggest to add some env var DUMP_ON_EXIT to give the possibility to control it from shell/docker-compose.

tianon:
This `trap` would only work during the duration of the temporary `initdb` server -- as soon as we `exec "$@"` at the end of the script, `mysqld` is the only thing running and this `trap` from the shell no longer exists.

Besides that, I'm not sure this functionality belongs in the base official image -- it's not really a generic feature that everyone wants/needs.

PR #680, opened 2020-05-29
Add Oracle Linux image variants

### LINKED ISSUE #289: Using OracleLinux as a base
This isn't a bug, more meant for discussion.

We've talked a bit about this before, but then there was less benefit to switching. With the new slim base images of OracleLinux, it may be of value to consider switching/adding variants, so I will just try to summarize pros and cons.

Pros:
* Image size is roughly halved. ol-slim is of comparable size to debian, while the custom Docker rpms on repo.mysql.com are much smaller than the standard Debian packages we use now. For 8.0 however there are the new -core packages for Debian that give the same size benefit (PR #282)
* Rpm package is more standardized across all major server versions. No need for the generic linux package used for 5.5

Cons:
* No apt. OracleLinux has yum. This is the big one, as it will break any image that uses FROM mysql, and then runs apt-anything.
* Risk of breaking other use cases that work on Debian but not rhel-based systems.

Possible ways to do it:
* Switch all: I don't really consider this an option, as such a major change to "stable" versions (even if most users of the image probably won't be affected) is not acceptable.
* Switch: One or a couple: The change is the biggest benefit for 5.5, since it would get rid of the generic linux package. But it would still be breaking someone's build. Also, having one version on OracleLinux and the rest on Debian will pretty much invert the "standardized" benefit :)
* Switch 8.0: Safer, since as a dmr it is not considered production ready, but we don't get the benefit since we have smaller 8.0.1 packages for deb as well. 
* Offer both: Best solution from user perspective, but it's already annoying and risky having to maintain three redundant copies of the Dockerfile and entrypoint script. Could we generate the files from a common source in some way?

The last option at least sounds best to me, but only if it's possible to have those files generated in some way, or if we could find a fairly clean way to have one script file handle all of them and just keep the copies fully in sync.

@tianon 
@yosifkit 
What are your thoughts on this?

### LINKED ISSUE #318: support arm64 architecture

### LINKED ISSUE #390: Simplify configuration files tree
A manual inspection to look for an option takes an unreasonable amount of time.

### LINKED ISSUE #822: Mysql Shell missing
Hi,

it would be nice to include support or the mysql shell for faster dumping and importing.

tianon:
The primary benefits of this are arm64 support (in 8.0) and smaller images (in 5.7 and 8.0).

```console
$ docker images mysql --format 'table {{ .Tag }}\t{{ .Size }}' | sort -rV
TAG          SIZE
8.0-oracle   404MB
8.0          556MB
5.7-oracle   371MB
5.7          447MB
5.6-oracle   386MB
5.6          303MB
```

Closes #390
Closes #318
Closes #289
Closes #822

This is a fusion of what @ltangvald started in https://github.com/docker-library/mysql/issues/289#issuecomment-547328690, the existing Debian variants, and what's in https://github.com/mysql/mysql-docker (in a very similar way to the layout @yosifkit described/suggested in https://github.com/docker-library/mysql/issues/289#issuecomment-548153826).

### DIFF
--- a/5.7/Dockerfile.oracle
+++ b/5.7/Dockerfile.oracle
@@ -1,0 +1,108 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
+FROM oraclelinux:7-slim
+
+RUN set -eux; \
+	groupadd --system --gid 999 mysql; \
+	useradd --system --uid 999 --gid 999 --home-dir /var/lib/mysql --no-create-home mysql; \
+	\
+	mkdir /var/lib/mysql /var/run/mysqld; \
+	chown mysql:mysql /var/lib/mysql /var/run/mysqld; \
+# ensure that /var/run/mysqld (used for socket and lock files) is writable regardless of the UID our mysqld instance ends up having at runtime
+	chmod 1777 /var/lib/mysql /var/run/mysqld; \
+	\
+	mkdir /docker-entrypoint-initdb.d
+
+# add gosu for easy step-down from root
+# https://github.com/tianon/gosu/releases
+ENV GOSU_VERSION 1.12
+RUN set -eux; \
+# TODO find a better userspace architecture detection method than querying the kernel
+	arch="$(uname -m)"; \
+	case "$arch" in \
+		aarch64) gosuArch='arm64' ;; \
+		x86_64) gosuArch='amd64' ;; \
+		*) echo >&2 "error: unsupported architecture: '$arch'"; exit 1 ;; \
+	esac; \
+	curl -fL -o /usr/local/bin/gosu.asc "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$gosuArch.asc"; \
+	curl -fL -o /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$gosuArch"; \
+	export GNUPGHOME="$(mktemp -d)"; \
+	gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4; \
+	gpg --batch --verify /usr/local/bin/gosu.asc /usr/local/bin/gosu; \
+	rm -rf "$GNUPGHOME" /usr/local/bin/gosu.asc; \
+	chmod +x /usr/local/bin/gosu; \
+	gosu --version; \
+	gosu nobody true
+
+RUN set -eux; \
+# https://dev.mysql.com/doc/refman/8.0/en/checking-gpg-signature.html
+# gpg: key 3A79BD29: public key "MySQL Release Engineering <mysql-build@oss.oracle.com>" imported
+	key='859BE8D7C586F538430B19C2467B942D3A79BD29'; \
+	export GNUPGHOME="$(mktemp -d)"; \
+	gpg --batch --keyserver keyserver.ubuntu.com --recv-keys "$key"; \
+	gpg --batch --export --armor "$key" > /etc/pki/rpm-gpg/RPM-GPG-KEY-mysql; \
+	rm -rf "$GNUPGHOME"
+
+ENV MYSQL_MAJOR 5.7
+ENV MYSQL_VERSION 5.7.37-1.el7
+
+RUN set -eu; \
+	. /etc/os-release; \
+	{ \
+		echo '[mysql5.7-server-minimal]'; \
+		echo 'name=MySQL 5.7 Server Minimal'; \
+		echo 'enabled=1'; \
+		echo "baseurl=https://repo.mysql.com/yum/mysql-5.7-community/docker/el/${VERSION_ID%%[.-]*}/\$basearch/"; \
+		echo 'gpgcheck=1'; \
+		echo 'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-mysql'; \
+# https://github.com/docker-library/mysql/pull/680#issuecomment-825930524
+		echo 'module_hotfixes=true'; \
+	} | tee /etc/yum.repos.d/mysql-community-minimal.repo
+
+RUN set -eux; \
+	yum install -y "mysql-community-server-minimal-$MYSQL_VERSION"; \
+	yum clean all; \
+# the "socket" value in the Oracle packages is set to "/var/lib/mysql" which isn't a great place for the socket (we want it in "/var/run/mysqld" instead)
+# https://github.com/docker-library/mysql/pull/680#issuecomment-636121520
+	grep -F 'socket=/var/lib/mysql/mysql.sock' /etc/my.cnf; \
+	sed -i 's!^socket=.*!socket=/var/run/mysqld/mysqld.sock!' /etc/my.cnf; \
+	grep -F 'socket=/var/run/mysqld/mysqld.sock' /etc/my.cnf; \
+	\
+# make sure users dumping files in "/etc/mysql/conf.d" still works
+	! grep -F '!includedir' /etc/my.cnf; \
+	{ echo; echo '!includedir /etc/mysql/conf.d/'; } >> /etc/my.cnf; \
+	mkdir -p /etc/mysql/conf.d; \
+	\
+	mysqld --version; \
+	mysql --version
+
+RUN set -eu; \
+	. /etc/os-release; \
+	{ \
+		echo '[mysql-tools-community]'; \
+		echo 'name=MySQL Tools Community'; \
+		echo "baseurl=https://repo.mysql.com/yum/mysql-tools-community/el/${VERSION_ID%%[.-]*}/\$basearch/"; \
+		echo 'enabled=1'; \
+		echo 'gpgcheck=1'; \
+		echo 'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-mysql'; \
+# https://github.com/docker-library/mysql/pull/680#issuecomment-825930524
+		echo 'module_hotfixes=true'; \
+	} | tee /etc/yum.repos.d/mysql-community-tools.repo
+ENV MYSQL_SHELL_VERSION 8.0.28-1.el7
+RUN set -eux; \
+	yum install -y "mysql-shell-$MYSQL_SHELL_VERSION"; \
+	yum clean all; \
+	\
+	mysqlsh --version
+
+VOLUME /var/lib/mysql
+
+COPY docker-entrypoint.sh /usr/local/bin/
+ENTRYPOINT ["docker-entrypoint.sh"]
+
+EXPOSE 3306 33060
+CMD ["mysqld"]
--- a/8.0/Dockerfile.oracle
+++ b/8.0/Dockerfile.oracle
@@ -1,0 +1,111 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
+FROM oraclelinux:8-slim
+
+RUN set -eux; \
+	groupadd --system --gid 999 mysql; \
+	useradd --system --uid 999 --gid 999 --home-dir /var/lib/mysql --no-create-home mysql; \
+	\
+	mkdir /var/lib/mysql /var/run/mysqld; \
+	chown mysql:mysql /var/lib/mysql /var/run/mysqld; \
+# ensure that /var/run/mysqld (used for socket and lock files) is writable regardless of the UID our mysqld instance ends up having at runtime
+	chmod 1777 /var/lib/mysql /var/run/mysqld; \
+	\
+	mkdir /docker-entrypoint-initdb.d
+
+# add gosu for easy step-down from root
+# https://github.com/tianon/gosu/releases
+ENV GOSU_VERSION 1.12
+RUN set -eux; \
+# TODO find a better userspace architecture detection method than querying the kernel
+	arch="$(uname -m)"; \
+	case "$arch" in \
+		aarch64) gosuArch='arm64' ;; \
+		x86_64) gosuArch='amd64' ;; \
+		*) echo >&2 "error: unsupported architecture: '$arch'"; exit 1 ;; \
+	esac; \
+	curl -fL -o /usr/local/bin/gosu.asc "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$gosuArch.asc"; \
+	curl -fL -o /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$gosuArch"; \
+	export GNUPGHOME="$(mktemp -d)"; \
+	gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4; \
+	gpg --batch --verify /usr/local/bin/gosu.asc /usr/local/bin/gosu; \
+	rm -rf "$GNUPGHOME" /usr/local/bin/gosu.asc; \
+	chmod +x /usr/local/bin/gosu; \
+	gosu --version; \
+	gosu nobody true
+
+RUN set -eux; \
+# https://dev.mysql.com/doc/refman/8.0/en/checking-gpg-signature.html
+# gpg: key 3A79BD29: public key "MySQL Release Engineering <mysql-build@oss.oracle.com>" imported
+	key='859BE8D7C586F538430B19C2467B942D3A79BD29'; \
+	export GNUPGHOME="$(mktemp -d)"; \
+	gpg --batch --keyserver keyserver.ubuntu.com --recv-keys "$key"; \
+	gpg --batch --export --armor "$key" > /etc/pki/rpm-gpg/RPM-GPG-KEY-mysql; \
+	rm -rf "$GNUPGHOME"
+
+# Oracle Linux 8+ is very slim :)
+RUN set -eux; microdnf install -y findutils; microdnf clean all
+
+ENV MYSQL_MAJOR 8.0
+ENV MYSQL_VERSION 8.0.28-1.el8
+
+RUN set -eu; \
+	. /etc/os-release; \
+	{ \
+		echo '[mysql8.0-server-minimal]'; \
+		echo 'name=MySQL 8.0 Server Minimal'; \
+		echo 'enabled=1'; \
+		echo "baseurl=https://repo.mysql.com/yum/mysql-8.0-community/docker/el/${VERSION_ID%%[.-]*}/\$basearch/"; \
+		echo 'gpgcheck=1'; \
+		echo 'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-mysql'; \
+# https://github.com/docker-library/mysql/pull/680#issuecomment-825930524
+		echo 'module_hotfixes=true'; \
+	} | tee /etc/yum.repos.d/mysql-community-minimal.repo
+
+RUN set -eux; \
+	microdnf install -y "mysql-community-server-minimal-$MYSQL_VERSION"; \
+	microdnf clean all; \
+# the "socket" value in the Oracle packages is set to "/var/lib/mysql" which isn't a great place for the socket (we want it in "/var/run/mysqld" instead)
+# https://github.com/docker-library/mysql/pull/680#issuecomment-636121520
+	grep -F 'socket=/var/lib/mysql/mysql.sock' /etc/my.cnf; \
+	sed -i 's!^socket=.*!socket=/var/run/mysqld/mysqld.sock!' /etc/my.cnf; \
+	grep -F 'socket=/var/run/mysqld/mysqld.sock' /etc/my.cnf; \
+	\
+# make sure users dumping files in "/etc/mysql/conf.d" still works
+	! grep -F '!includedir' /etc/my.cnf; \
+	{ echo; echo '!includedir /etc/mysql/conf.d/'; } >> /etc/my.cnf; \
+	mkdir -p /etc/mysql/conf.d; \
+	\
+	mysqld --version; \
+	mysql --version
+
+RUN set -eu; \
+	. /etc/os-release; \
+	{ \
+		echo '[mysql-tools-community]'; \
+		echo 'name=MySQL Tools Community'; \
+		echo "baseurl=https://repo.mysql.com/yum/mysql-tools-community/el/${VERSION_ID%%[.-]*}/\$basearch/"; \
+		echo 'enabled=1'; \
+		echo 'gpgcheck=1'; \
+		echo 'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-mysql'; \
+# https://github.com/docker-library/mysql/pull/680#issuecomment-825930524
+		echo 'module_hotfixes=true'; \
+	} | tee /etc/yum.repos.d/mysql-community-tools.repo
+ENV MYSQL_SHELL_VERSION 8.0.28-1.el8
+RUN set -eux; \
+	microdnf install -y "mysql-shell-$MYSQL_SHELL_VERSION"; \
+	microdnf clean all; \
+	\
+	mysqlsh --version
+
+VOLUME /var/lib/mysql
+
+COPY docker-entrypoint.sh /usr/local/bin/
+ENTRYPOINT ["docker-entrypoint.sh"]
+
+EXPOSE 3306 33060
+CMD ["mysqld"]
--- a/apply-templates.sh
+++ b/apply-templates.sh
@@ -31,15 +31,18 @@ generated_warning() {
 for version; do
 	export version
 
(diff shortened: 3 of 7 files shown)


tianon:
Here are the configuration differences from the existing Debian variants to these new Oracle Linux variants:

# 8.0

```diff
$ diff -u <(docker run --rm --hostname test mysql:8.0 mysqld --verbose --help) <(docker run --rm --hostname test mysql:8.0-oracle mysqld --verbose --help)
--- /dev/fd/63	2020-05-29 11:22:08.853724508 -0700
+++ /dev/fd/62	2020-05-29 11:22:08.853724508 -0700
@@ -10,7 +10,7 @@
 Usage: mysqld [OPTIONS]
 
 Default options are read from the following files in the given order:
-/etc/my.cnf /etc/mysql/my.cnf ~/.my.cnf 
+/etc/my.cnf /etc/mysql/my.cnf /usr/etc/my.cnf ~/.my.cnf 
 The following groups are read: mysqld server mysqld-8.0
 The following options may be given as the first argument:
 --print-defaults        Print the program argument list and exit.
@@ -757,9 +757,6 @@
                       Reset a monitor counter
   --innodb-monitor-reset-all=name 
                       Reset all values for a monitor counter
-  --innodb-numa-interleave 
-                      Use NUMA interleave memory policy to allocate InnoDB
-                      buffer pool.
   --innodb-old-blocks-pct=# 
                       Percentage of the buffer pool to reserve for 'old'
                       blocks.
@@ -2209,7 +2206,6 @@
 innodb-monitor-enable                                        (No default value)
 innodb-monitor-reset                                         (No default value)
 innodb-monitor-reset-all                                     (No default value)
-innodb-numa-interleave                                       FALSE
 innodb-old-blocks-pct                                        37
 innodb-old-blocks-time                                       1000
 innodb-online-alter-log-max-size                             134217728
@@ -2465,7 +2461,7 @@
 persist-only-admin-x509-subject                              
 persisted-globals-load                                       TRUE
 pid-file                                                     /var/run/mysqld/mysqld.pid
-plugin-dir                                                   /usr/lib/mysql/plugin/
+plugin-dir                                                   /usr/lib64/mysql/plugin/
 port                                                         3306
 port-open-timeout                                            0
 preload-buffer-size                                          32768
@@ -2499,7 +2495,7 @@
 safe-user-create                                             FALSE
 schema-definition-cache                                      256
 secondary-engine-cost-threshold                              100000
-secure-file-priv                                             NULL
+secure-file-priv                                             /var/lib/mysql-files
 server-id                                                    1
 server-id-bits                                               32
 session-track-gtids                                          OFF
@@ -2539,7 +2535,7 @@
 slow-launch-time                                             2
 slow-query-log                                               FALSE
 slow-query-log-file                                          /var/lib/mysql/test-slow.log
-socket                                                       /var/run/mysqld/mysqld.sock
+socket                                                       /var/lib/mysql/mysql.sock
 sort-buffer-size                                             262144
 sporadic-binlog-dump-fail                                    FALSE
 sql-mode                                                     ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION
@@ -2575,7 +2571,7 @@
 thread-handling                                              one-thread-per-connection
 thread-stack                                                 286720
 tls-ciphersuites                                             (No default value)
-tls-version                                                  TLSv1,TLSv1.1,TLSv1.2,TLSv1.3
+tls-version                                                  TLSv1,TLSv1.1,TLSv1.2
 tmp-table-size                                               16777216
 tmpdir                                                       /tmp
 transaction-alloc-block-size                                 8192
```

# 5.7

```diff
$ diff -u <(docker run --rm --hostname test mysql:5.7 mysqld --verbose --help) <(docker run --rm --hostname test mysql:5.7-oracle mysqld --verbose --help)
--- /dev/fd/63	2020-05-29 11:22:39.796833194 -0700
+++ /dev/fd/62	2020-05-29 11:22:39.800833080 -0700
@@ -10,7 +10,7 @@
 Usage: mysqld [OPTIONS]
 
 Default options are read from the following files in the given order:
-/etc/my.cnf /etc/mysql/my.cnf ~/.my.cnf 
+/etc/my.cnf /etc/mysql/my.cnf /usr/etc/my.cnf ~/.my.cnf 
 The following groups are read: mysqld server mysqld-5.7
 The following options may be given as the first argument:
 --print-defaults        Print the program argument list and exit.
@@ -2149,7 +2149,7 @@
 performance-schema-setup-objects-size                        -1
 performance-schema-users-size                                -1
 pid-file                                                     /var/run/mysqld/mysqld.pid
-plugin-dir                                                   /usr/lib/mysql/plugin/
+plugin-dir                                                   /usr/lib64/mysql/plugin/
 port                                                         3306
 port-open-timeout                                            0
 preload-buffer-size                                          32768
@@ -2223,7 +2223,7 @@
 slow-launch-time                                             2
 slow-query-log                                               FALSE
 slow-query-log-file                                          /var/lib/mysql/test-slow.log
-socket                                                       /var/run/mysqld/mysqld.sock
+socket                                                       /var/lib/mysql/mysql.sock
 sort-buffer-size                                             262144
 sporadic-binlog-dump-fail                                    FALSE
 sql-mode                                                     ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
```

# 5.6

```diff
$ diff -u <(docker run --rm --hostname test mysql:5.6 mysqld --verbose --help) <(docker run --rm --hostname test mysql:5.6-oracle mysqld --verbose --help)
2020-05-29 18:22:59 0 [Note] mysqld (mysqld 5.6.48) starting as process 1 ...
2020-05-29 18:22:59 1 [Note] Plugin 'FEDERATED' is disabled.
mysqld: Table 'mysql.plugin' doesn't exist
2020-05-29 18:22:59 1 [ERROR] Can't open the mysql.plugin table. Please run mysql_upgrade to create it.
2020-05-29 18:22:59 0 [Note] mysqld (mysqld 5.6.48) starting as process 1 ...
2020-05-29 18:22:59 1 [Note] Plugin 'FEDERATED' is disabled.
mysqld: Table 'mysql.plugin' doesn't exist
2020-05-29 18:22:59 1 [ERROR] Can't open the mysql.plugin table. Please run mysql_upgrade to create it.
2020-05-29 18:22:59 1 [Note] Binlog end
2020-05-29 18:22:59 1 [Note] Shutting down plugin 'MyISAM'
2020-05-29 18:22:59 1 [Note] Shutting down plugin 'CSV'
2020-05-29 18:22:59 1 [Note] Binlog end
2020-05-29 18:22:59 1 [Note] Shutting down plugin 'MyISAM'
2020-05-29 18:22:59 1 [Note] Shutting down plugin 'CSV'
--- /dev/fd/63	2020-05-29 11:22:59.296271443 -0700
+++ /dev/fd/62	2020-05-29 11:22:59.296271443 -0700
@@ -10,7 +10,7 @@
 Usage: mysqld [OPTIONS]
 
 Default options are read from the following files in the given order:
-/etc/my.cnf /etc/mysql/my.cnf ~/.my.cnf 
+/etc/my.cnf /etc/mysql/my.cnf /usr/etc/my.cnf ~/.my.cnf 
 The following groups are read: mysqld server mysqld-5.6
 The following options may be given as the first argument:
 --print-defaults        Print the program argument list and exit.
@@ -1934,7 +1934,7 @@
 performance-schema-setup-objects-size                      100
 performance-schema-users-size                              -1
 pid-file                                                   /var/run/mysqld/mysqld.pid
-plugin-dir                                                 /usr/lib/mysql/plugin/
+plugin-dir                                                 /usr/lib64/mysql/plugin/
 port                                                       3306
 port-open-timeout                                          0
 preload-buffer-size                                        32768
@@ -1996,7 +1996,7 @@
 slow-launch-time                                           2
 slow-query-log                                             FALSE
 slow-query-log-file                                        /var/lib/mysql/test-slow.log
-socket                                                     /var/run/mysqld/mysqld.sock
+socket                                                     /var/lib/mysql/mysql.sock
 sort-buffer-size                                           262144
 sporadic-binlog-dump-fail                                  FALSE
 sql-mode                                                   NO_ENGINE_SUBSTITUTION
```

tianon:
Of those changes, this is the one that concerns me from a usability perspective: (since I can imagine we have users relying on the default value of `socket`)

```diff
-socket                                                 /var/run/mysqld/mysqld.sock
+socket                                                 /var/lib/mysql/mysql.sock
```

Also, I'm a little confused about the missing `innodb-numa-interleave` (which seems strange) and TLSv1.3 support (which is actually somewhat alarming) in MySQL 8.0 in these RPMs.

Hopefully @ltangvald has thoughts / opinions / time to review? :pray: :heart: :+1:

thaJeztah:
could a symlink work for the socket?

tianon:
I don't think so -- the use case I'm worried about is users who have created a volume on `/var/run/mysqld` (which would overwrite our symlink anyhow) explicitly to share with the host or other containers, but for whom sharing `/var/lib/mysql` is not acceptable (given that it contains the actual database data :sweat_smile:).

thaJeztah:
Gotcha! Yes it's a bit of an odd location for a socket 🤔

ltangvald:
Will try to work through one thing at a time, since I haven't been able to focus on Docker in some time, so forgive me if I ramble a bit in the comments :)

For the default socket location, I don't think there's any good reason to use var/lib/mysql aside from historical ones (on rhel systems). It will actually cause problems on e.g. Windows, because windows won't let docker map sockets to a bind-mounted volume location (I'm not 100% sure this is still the case, but we mentioned it at https://dev.mysql.com/doc/mysql-installation-excerpt/8.0/en/deploy-mysql-nonlinux-docker.html)

So my suggestion is to override this setting to be the same as what we do for Debian, i.e. {var}/run/mysqld/mysqld.sock

ltangvald:
For tls support, we build with system openssl (where possible), which on OL7 is 1.0.2, which I believe only supports 1.2. OL8 has 1.1.1, but 8-slim doesn't actually have yum or even dnf by default (only microdnf, which isn't 1:1), so it would mean some changes.

ltangvald:
The missing numa support is probably because the build didn't have access to an installed numa package, but I'll have to look into the cause of it Might be down to ol7 again, but should at least be something we'll get back in down the line.

ltangvald:
We should maybe just use the same config files we use the debian-based 8.0 image? Though the rpm defaults to installing /etc/my.cnf instead of /etc/mysql/my.cnf

tianon:
> We should maybe just use the same config files we use the debian-based 8.0 image? Though the rpm defaults to installing /etc/my.cnf instead of /etc/mysql/my.cnf

Yeah, I think this is one of those cases where users should maybe understand/accept the differences in convention?  We don't want to force too many Debian-isms on the Oracle-based images, IMO.

tianon:
(also, I've finally updated this to the latest version, rebased, added the `socket=...` fixes)

tianon:
This is now successfully rebased on top of #730. :muscle:

tianon:
(@Djelibeybi I think I've already shared this PR with you previously, but it's likely relevant to your interests :eyes:)

Djelibeybi:
@tianon thanks, it certainly is. 

Question for you and @ltangvald: is there a reason you went with `oraclelinux:7-slim` instead of `oraclelinux:8-slim`? We're up to Update 3 now, it's pretty stable. :)

tianon:
Heh, that's merely a factor of the PR's age!  I'll take a look at updating to 8 instead. :+1:

Djelibeybi:
I figured it might be. There's still a number of years left on OL7, but having both would be nice. :)

tianon:
Aha, now I see it:

- https://repo.mysql.com/yum/mysql-8.0-community/docker/x86_64/
- https://repo.mysql.com/yum/mysql-5.7-community/docker/x86_64/
- https://repo.mysql.com/yum/mysql-5.6-community/docker/x86_64/

All only published for `el7` :innocent:

Djelibeybi:
Well, that certainly makes it more challenging. @ltangvald any plans for OL8 RPMs? 😄

tianon:
> Well, that certainly makes it more challenging. @ltangvald any plans for OL8 RPMs?

@Djelibeybi in the meantime, we should move forward with the ol7 variants? :smile:

tianon:
(It's not like upgrading to ol8 later will bring many breaking changes for the way the image is used :smile:)

ltangvald:
I actually started working on this again recently, and ran into the ol8 issue. We do have minimal packages in the works, but there is an issue with the ol8-slim base image only having microdnf, which is not 1:1 with yum. We can install dnf, but that pulls in so much extra stuff we kind of lose the advantage of the slim image to begin with.

So for now, maybe best if I just try to complete this with ol7 so we actually have some arm images (the biggest disadvantage is that ol7 has an old ssl, I think?), and we can look at migrating to ol8 when we work out a nice way of doing it upstream?

Djelibeybi:
> there is an issue with the ol8-slim base image only having microdnf, which is not 1:1 with yum. 

Ping me internally and we can resolve any issues you're having with `microdnf`. There should be no reason to require the full `dnf` client.

Djelibeybi:
But @tianon, I agree with @ltangvald that we should continue with the OL7 version for now.

Djelibeybi:
> (It's not like upgrading to ol8 later will bring many breaking changes for the way the image is used 😄)

Actually, it does. Using `oraclelinux:7-slim` means that `yum` is available to downstream images. If we use `oraclelinux:8-slim` then neither `yum` nor even `dnf` are available, just `microdnf`.

I'm tempted to suggest a larger `oraclelinux:8` based option (for folks who want downstream flexibility) and an `oraclelinux:8-slim` option for those who just want the smallest possible image size.

tianon:
Ok, I'll bite - what's the use case for building an image `FROM mysql:xxx` and installing more things? :innocent:

Djelibeybi:
> Ok, I'll bite - what's the use case for building an image `FROM mysql:xxx` and installing more things? 😇

It was the reason given for not switching the MySQL image across to Oracle Linux and our RPMs in the first place, so 🤷 .

ltangvald:
We do get some requests for features upstream we don't really want to add (because it's not something we think most people want and it would add complexity), and then we'd instead ask them to extend the image themselves. I couldn't really say how many do this, though.

tianon:
I went to go rebase this, and noticed that http://repo.mysql.com/yum/mysql-8.0-community/docker/ includes 8.0.23 for el7, but 8.0.24 for el8, so I figured I'd give it a shot and update to OL8 for 8.0, but I'm seeing some strange behavior and feeling like I'm in way over my head. :innocent:

I've got the `/etc/yum.repos.d/mysql-community-minimal.repo` as in this PR, and `microdnf repolist` shows it properly:

```console
bash-4.4# microdnf repolist
repo id                                                           repo name                                                                       
mysql8.0-server-minimal                                           MySQL 8.0 Server Minimal                                                        
ol8_appstream                                                     Oracle Linux 8 Application Stream (x86_64)                                      
ol8_baseos_latest                                                 Oracle Linux 8 BaseOS Latest (x86_64)                                           
```

However, when I try to install `mysql-community-server-minimal` (which I've even dug into the XML repo data files to verify exists), `microdnf` disagrees.  The place it gets really weird for me is that it _can_ find (and successfully install!) `mysql-community-server-minimal-debuginfo-8.0.24-1.el8.x86_64`, which comes from the same repo, and corresponds to the package I actually want??

```console
bash-4.4# microdnf repoquery 'mysql-community-server*'
mysql-community-server-minimal-debuginfo-8.0.24-1.el8.x86_64
bash-4.4# microdnf install mysql-community-server-minimal
error: No package matches 'mysql-community-server-minimal'
bash-4.4# microdnf install mysql-community-server-minimal-debuginfo-8.0.24-1.el8.x86_64
Package                                                                                                Repository                             Size
Installing:                                                                                                                                       
 mysql-community-minimal-debugsource-8.0.24-1.el8.x86_64                                               mysql8.0-server-minimal             18.3 MB
 mysql-community-server-minimal-debuginfo-8.0.24-1.el8.x86_64                                          mysql8.0-server-minimal            250.9 MB
Transaction Summary:
 Installing:        2 packages
 Reinstalling:      0 packages
 Upgrading:         0 packages
 Removing:          0 packages
 Downgrading:       0 packages
Downloading packages...
Running transaction test...
Installing: mysql-community-minimal-debugsource;8.0.24-1.el8;x86_64;mysql8.0-server-minimal
Installing: mysql-community-server-minimal-debuginfo;8.0.24-1.el8;x86_64;mysql8.0-server-minimal
Complete.
```

tianon:
... looks like I caught the mirror in the middle of a transition from https://repo.mysql.com/yum/mysql-8.0-community/docker/ to https://repo.mysql.com/yum/mysql-8.0-community/docker/el/X/ :joy:

tianon:
(Although I'm seeing the same weird behavior with `http://repo.mysql.com/yum/mysql-8.0-community/docker/el/8/$basearch/`, so maybe there's still something else wrong or in-progress that I need to wait for. :bow:)

tianon:
```console
bash-4.4# microdnf --disablerepo=ol8_appstream --disablerepo=ol8_baseos_latest repoquery 'mysql-community*'
mysql-community-minimal-debugsource-8.0.24-1.el8.x86_64
mysql-community-server-minimal-8.0.24-1.el8.x86_64
mysql-community-server-minimal-debuginfo-8.0.24-1.el8.x86_64
```

:thinking: :confused:

(Disabling the OL8 repos makes it show up??)

tianon:
(To be clear, I'm not _happy_ with the workaround I just committed that amounts to the cheeky nonsense in my second edit to https://github.com/docker-library/mysql/pull/680#issuecomment-825906122, but it let me successfully finish this rebase for now and hopefully in a couple days it won't be the weekend for my Oracle friends and they'll help me dig myself out of whatever this hole is. :smile:)

Djelibeybi:
It's caused by modularity filtering out the packages in a non-modular channel. If you use the `oraclelinux:8` image (which has the full `dnf` client), this becomes more obvious. However, if you add `module_hotfixes=true` to each of the MySQL repos, it will disable this filtering. It's still a workaround, but less of a clunky one.

tianon:
Ohhhhh, interesting (and strange) -- thank you!!

(You are not obligated to respond to anything I ask of you, least of all on your Saturday, so please know that I appreciate it and hope it was because you found my floundering interesting. :smile:)

Djelibeybi:
Modularity is weird: https://dnf.readthedocs.io/en/latest/modularity.html

I can see what they're trying to achieve, and like `systemd`, I'll drink the Flavor-Aid, but it's just not quite there yet, IMO.

Djelibeybi:
FYI, the MySQL repo URLs can/should use `https://`.

Djelibeybi:
Also, the container built on `oraclelinux:8-slim` is only 64MB smaller than the one built on `oraclelinux:8` which has things like `findutils` and the full `dnf` client. Just food for thought.

tianon:
Iiiiiinteresting, although the layer that installs `findutils` is only 9.61MB :smile:

Djelibeybi:
Yes, that's one of the differences. Having the full `dnf` client lets you do things like `dnf -y config-manager --save --setopt=ol8_MySQL80.module_hotfixes=true` instead of hand-editing the file.

Djelibeybi:
You could even `dnf -y install mysql-release-el8` and then use `dnf config-manager --save --setopt=ol8_MySQL80.baseurl=https://repo.mysql.com/yum/mysql-8.0-community/docker/el/\$releasever/\$arch/`

tianon:
Heh, but all that seems a lot more complex than simply creating the `.repo` file with exactly the content we want as this PR does now? :innocent:

ltangvald:
@tianon We've finally got our 8.0.24 docker images out (had some pretty big reworking of how it's structured, so it's been some work). You can see how we do it now at https://github.com/mysql/mysql-docker (specifically https://github.com/mysql/mysql-docker/blob/main/mysql-cluster/8.0/Dockerfile)

tianon:
Sorry @ltangvald, I let this slip again :bow:

I've rebased to include the removal of 5.6 (and a few more updates that have happened since I last looked at this).

I've also looked at the `Dockerfile` you linked - this is helpful, thanks!  I'll try to explain a bit why I'm not just using it as-is though.

The primary concern I have is `rpm -U https://...` downloads/installs those RPMs without any verification; to add some kind of verification, I think we'd have to install a few more deps first so we can either do PGP or a checksum (or ideally both), and at that point I'm not sure we're adding value over just creating the config and installing the appropriate key as I've got in this PR? :see_no_evil:

We've also got the full `docker-entrypoint.sh` support from this repo to consider, although that's not much more (adding `gosu` and the appropriate script).

Beyond that, it's mostly just reorganizing things slightly for better cache when MySQL versions change. :sweat_smile:  Functionally, I think we're at a pretty similar place? :innocent:

Here are updated numbers from my latest build:

```console
$ docker images mysql --format 'table {{ .Tag }}\t{{ .Size }}' | sort -rV
TAG          SIZE
8.0-oracle   431MB
8.0-debian   519MB
5.7-oracle   395MB
5.7-debian   448MB
```

For @Djelibeybi's sake, I've also done a test build against `oraclelinux:8` instead of slim, for which I had to change the MySQL UID/GID (since the non-slim has more users and thus 999 is already taken) and swap `microdnf` to full `dnf`:

```console
$ docker images mysql:8.0-oracle-notslim --format 'table {{ .Tag }}\t{{ .Size }}'
TAG                  SIZE
8.0-oracle-notslim   563MB
```

ltangvald:
@tianon No worries! I had hoped to spend more time on this than I have, too :|

I agree with your points about the config package

For the config changes you do, maybe we should try to get it fully in line with the Debian image? Copy config/conf.d to /etc/mysql/ and config/my.cnf to /etc/?

tianon:
(I don't feel super strongly though, so if you still feel strongly I'll adjust this to copy that conf directory instead!)

ltangvald:
I was thinking more about the setup with my.cnf including files in /etc/mysql/conf.d, to keep a consistent way for users to add custom config files.

tianon:
Ahhhhh, I see what you mean -- my `--verbose --help` doesn't show `!includedir` directives so we don't see that we've lost `/etc/mysql/conf.d/` in the Oracle versions.  I'm tracking now.

tianon:
Ok, I've updated to add `!includedir /etc/mysql/conf.d/` at the end of `/etc/my.cnf` -- I noticed the 5.7 build ends up including `/etc/my.cnf.d`.  Should we be dropping overrides in there instead, or do you think modifying `my.cnf` is good enough?

ltangvald:
That seems odd at first glance. The config coming with the 5.7 minimal rpm shouldn't have that include line, I think. Redhat's own MySQL packages do, and in el8 we have the disablerepo=ol8_appstream flag to prevent that from interfering with the installation, but that shouldn't be an issue in el7, I 'd think.

tianon:
```console
$ docker run --rm mysql:5.7-oracle sh -c 'ls -ld /etc/my.cnf*'
-rw-r--r-- 1 root root 1045 Feb 15 17:39 /etc/my.cnf
drwxr-xr-x 2 root root 4096 Nov 29 20:31 /etc/my.cnf.d
$ docker run --rm mysql:5.7-oracle cat /etc/my.cnf
# For advice on how to change settings please see
# http://dev.mysql.com/doc/refman/5.7/en/server-configuration-defaults.html

[mysqld]
#
# Remove leading # and set to the amount of RAM for the most important data
# cache in MySQL. Start at 70% of total RAM for dedicated server, else 10%.
# innodb_buffer_pool_size = 128M
#
# Remove leading # to turn on a very important data integrity option: logging
# changes to the binary log between backups.
# log_bin
#
# Remove leading # to set options mainly useful for reporting servers.
# The server defaults are faster for transactions and fast SELECTs.
# Adjust sizes as needed, experiment to find the optimal values.
# join_buffer_size = 128M
# sort_buffer_size = 2M
# read_rnd_buffer_size = 2M
skip-host-cache
skip-name-resolve
datadir=/var/lib/mysql
socket=/var/run/mysqld/mysqld.sock
secure-file-priv=/var/lib/mysql-files
user=mysql

# Disabling symbolic-links is recommended to prevent assorted security risks
symbolic-links=0

log-error=/var/log/mysqld.log
pid-file=/var/run/mysqld/mysqld.pid
```

yosifkit:
This looks good to me.

@ltangvald are we good to merge this now? Any concerns or questions?

ltangvald:
@yosifkit @tianon Yeah, I think so. I've had several stabs at this before and then let myself get bogged down in details, which just results in it not getting done :)

tianon:
Well then, bombs away! :smile: :heart:

Let's get this into the hands of users so they can find issues we need to fix. :innocent:

rfay:
I don't see 5.7-oracle out there in arm64 at this time, just amd64. 

I'm sad to see these are oraclelinux instead of debian-derived.

rfay:
Yes, I understand that, but Ubuntu packaging system has arm64 packages for both. And of course those oraclelinux packages are completely different from everything else that has ever been published in this channel. I'm just sad, it's not a very good solution.

PR #724, opened 2020-10-21
no flush privs needed after grant/create user

yosifkit:
Mirroring that change in https://github.com/docker-library/mariadb/pull/296


tianon:
LGTM :+1:

PR #728, opened 2020-11-14
Execute SQL in entrypoint with "--comments"

### LINKED ISSUE #718: Lost SQL comments in /docker-entrypoint-initdb.d
Hi,

We do really like this image and use it to manage our migrations and run tests against a development image based on this one with the full schema and test data dumped to `/docker-entrypoint-initdb.d`.

The problem we encounter is that after restoring our schema we will loose all SQL comments we have defined in triggers, procedures, functions, ... which is quite annoying as this is our reference for development and we need our comments in complicated DB logic to be still available for further development.

The comments get lost in `docker_process_sql()` on:
```
mysql --defaults-file=<( _mysql_passfile "${passfileArgs[@]}") --protocol=socket -uroot -hlocalhost --socket="${SOCKET}" "$@"
```

This command is missing the option:
```
-c, --comments      Preserve comments. Send comments to the server. The
                    default is --skip-comments (discard comments), enable
                    with --comments.
```

I have simply added `--comments` and it works fine for my use case.

Please could you do so, too or at least provide a way to configure this in?

Thanks

marc-mabe:
... to prevent lost of SQL comments in triggers, procedures etc. on initdb.

fixes #718

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -208,7 +208,7 @@ docker_process_sql() {
 		set -- --database="$MYSQL_DATABASE" "$@"
 	fi
 
-	mysql --defaults-extra-file=<( _mysql_passfile "${passfileArgs[@]}") --protocol=socket -uroot -hlocalhost --socket="${SOCKET}" "$@"
+	mysql --defaults-extra-file=<( _mysql_passfile "${passfileArgs[@]}") --protocol=socket -uroot -hlocalhost --socket="${SOCKET}" --comments "$@"
 }
 
 # Initializes database with timezone info and root password, plus optional extra db/user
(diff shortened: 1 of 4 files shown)


tianon:
Rebased, updated the template copy also, and added this note to the commit message:

> The primary reason to disable this seems to be to avoid normally-unnecessary data transfer, but in this case we're explicitly connecting over the local unix socket, so streaming the comments to the server and letting it strip them should be *almost* as efficient as the client doing so.

tianon:
Thank you for the contribution (and your patience)! :heart:

slashformotion:
For reference, this does not work as intended for MySQL 5.7.41 and we had to remove it by using 
```sh
RUN sed -i 's/--comments //' /usr/local/bin/docker-entrypoint.sh 
````
in our Dockerfile. The issue was the comment at the top of the file made the import of the MySQL dump crash.

PR #734, opened 2020-12-11
Revert versions to standard Dockerfile path

beornf:
This helps to make it easier to work with standard Docker tool defaults. If variations were added later (such as alpine) its more consistent to split these variations into top-level directories.


tianon:
Sorry for the delay here -- I understand the motivation, but this was unfortunately a deliberate decision to make the Oracle/Debian split more clear post-#680 (especially since in whatever major release comes after 8.0, Oracle will be the default instead of Debian, which would make a bare `Dockerfile` more confusing and more complicated in our templating). :pray: :heart:

Issue #543, opened 2019-02-26
Using option `--default-time-zone` at first run makes `mysql_install_db` script fail


yannoff:
When using option `--default-time-zone` in docker command, the MySQL initialization script fails, because the timezone tables haven't been populated yet when `mysql_install_db` is invoked.

Here is an example, using the following `docker-compose` configuration:

```yaml
#docker-compose.yml
version: '3'
services:
    mysql:
        image: mysql:5.5
        environment:
            MYSQL_DATABASE: acme
            MYSQL_ROOT_PASSWORD: xxxyyy
        command:
             - mysqld
             - "--character-set-server=utf8mb4"
             - "--skip-character-set-client-handshake"
             - "--collation-server=utf8mb4_unicode_ci"
             - "--default-time-zone=Europe/Paris"
```

And here is the resulting log:

```log
Attaching to webapp_mysql_1
mysql_1        | Initializing database
mysql_1        | 190219 11:01:18 [Note] Ignoring --secure-file-priv value as server is running with --bootstrap.
mysql_1        | 190219 11:01:18 [Note] /usr/local/mysql/bin/mysqld (mysqld 5.5.62) starting as process 68 ...
mysql_1        | 190219 11:01:18 [ERROR] Fatal error: Illegal or unknown default time zone 'Europe/Paris'
mysql_1        | 
mysql_1        | Installation of system tables failed!  Examine the logs in
mysql_1        | /var/lib/mysql/ for more information.
mysql_1        | 
mysql_1        | You can try to start the mysqld daemon with:
mysql_1        | 
mysql_1        |     shell> /usr/local/mysql/bin/mysqld --skip-grant &
mysql_1        | 
mysql_1        | and use the command line tool /usr/local/mysql/bin/mysql
mysql_1        | to connect to the mysql database and look at the grant tables:
mysql_1        | 
mysql_1        |     shell> /usr/local/mysql/bin/mysql -u root mysql
mysql_1        |     mysql> show tables
mysql_1        | 
mysql_1        | Try 'mysqld --help' if you have problems with paths.  Using --log
mysql_1        | gives you a log in /var/lib/mysql/ that may be helpful.
mysql_1        | 
mysql_1        | Please consult the MySQL manual section
mysql_1        | 'Problems running mysql_install_db', and the manual section that
mysql_1        | describes problems on your OS.  Another information source are the
mysql_1        | MySQL email archives available at http://lists.mysql.com/.
mysql_1        | 
[...]

PR #739, opened 2021-02-10
Override time-zone option when initializing database

glensc:
Initializing a database with a config that specifies `default-time-zone` ends up with an error:

> Fatal error: Illegal or unknown default time zone 'Europe/Tallinn'

This adds a workaround to the problem to force the system timezone (similarly to `--initialize-insecure`, `--skip-networking` options used in init db process), so the database can be initialized, and timezones be imported after database init.

Implements solution described at:
- https://github.com/docker-library/mysql/issues/543#issuecomment-775895366

Why this is brought up again? It's because a config file offers a unified way to configure MySQL instance, compared to setting up environment variables for different systems (docker run, docker-compose, kubernetes, ...).

---- 

Local build commands:
```
git clone --depth 1 -b patch-1 https://github.com/glensc/docker-library-mysql mysql-pr739 && cd mysql-pr739
docker build 5.6 -t mysql:5.6_pr739 -f 5.6/Dockerfile.debian
docker build 5.7 -t mysql:5.7_pr739 -f 5.7/Dockerfile.debian
docker build 8.0 -t mysql:8.0_pr739 -f 8.0/Dockerfile.debian
```

Local test commands:

```
docker run --rm -e MYSQL_ALLOW_EMPTY_PASSWORD=1 mysql:5.6_pr739 gosu mysql bash -x /entrypoint.sh mysqld --default-time-zone=Europe/Tallinn
docker run --rm -e MYSQL_ALLOW_EMPTY_PASSWORD=1 mysql:5.7_pr739 gosu mysql bash -x /entrypoint.sh mysqld --default-time-zone=Europe/Tallinn
docker run --rm -e MYSQL_ALLOW_EMPTY_PASSWORD=1 mysql:8.0_pr739 gosu mysql bash -x /entrypoint.sh mysqld --default-time-zone=Europe/Tallinn
```

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -98,7 +98,7 @@ mysql_get_config() {
 # Do a temporary startup of the MySQL server, for init purposes
 docker_temp_server_start() {
 	if [ "${MYSQL_MAJOR}" = '5.6' ] || [ "${MYSQL_MAJOR}" = '5.7' ]; then
-		"$@" --skip-networking --socket="${SOCKET}" &
+		"$@" --skip-networking --default-time-zone=SYSTEM --socket="${SOCKET}" &
 		mysql_note "Waiting for server startup"
 		local i
 		for i in {30..0}; do
@@ -118,7 +118,7 @@ docker_temp_server_start() {
 		fi
 	else
 		# For 5.7+ the server is ready for use as soon as startup command unblocks
-		if ! "$@" --daemonize --skip-networking --socket="${SOCKET}"; then
+		if ! "$@" --daemonize --skip-networking --default-time-zone=SYSTEM --socket="${SOCKET}"; then
 			mysql_error "Unable to start server."
 		fi
 	fi
@@ -158,9 +158,9 @@ docker_create_db_directories() {
 docker_init_database_dir() {
 	mysql_note "Initializing database files"
 	if [ "$MYSQL_MAJOR" = '5.6' ]; then
-		mysql_install_db --datadir="$DATADIR" --rpm --keep-my-cnf "${@:2}"
+		mysql_install_db --datadir="$DATADIR" --rpm --keep-my-cnf "${@:2}" --default-time-zone=SYSTEM
 	else
-		"$@" --initialize-insecure
+		"$@" --initialize-insecure --default-time-zone=SYSTEM
 	fi
 	mysql_note "Database files initialized"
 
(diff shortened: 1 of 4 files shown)


yannoff:
IMHO https://github.com/docker-library/mysql/blob/master/template/docker-entrypoint.sh would be the only file to be changed, seems like the distributed `Dockerfiles` are automatically generated then... ping @tianon ?

glensc:
> IMHO [https://github.com/docker-library/mysql/blob/master/template/docker-entrypoint.sh](https://github.com/docker-library/mysql/blob/master/template/docker-entrypoint.sh?rgh-link-date=2021-02-10T22%3A23%3A18Z) ([ hace 20 minutos](https://github.com/docker-library/mysql/blob/c273193c14bb45daea2d27bb6ec51aa0b2b2ab9b/template/docker-entrypoint.sh?rgh-link-date=2021-02-10T22%3A23%3A18Z)) would be the only file to be changed, seems like the distributed `Dockerfile`s are automatically generated then...

@yannoff that's exactly what I said to you:
- https://github.com/docker-library/mysql/issues/543#issuecomment-777069712

glensc:
@yannoff 

🚧 Never link to branches, share a permalink 🚧
- see https://news.ycombinator.com/item?id=8046710
  and https://help.github.com/articles/getting-permanent-links-to-files/
  and https://docs.gitlab.com/ee/user/shortcuts.html#project-files
- hit `y`
- and perhaps for https://github.com/josephfrazier/octopermalinker (GitHub)

NOTE: edit your post and fix it with a permalink!

yannoff:
> > IMHO [https://github.com/docker-library/mysql/blob/master/template/docker-entrypoint.sh](https://github.com/docker-library/mysql/blob/master@%7B2021-02-10T22:44:33Z%7D/template/docker-entrypoint.sh) would be the only file to be changed, seems like the distributed `Dockerfile`s are automatically generated then...
> 
> @yannoff that's exactly what I said to you:
> 
>     * [#543 (comment)](https://github.com/docker-library/mysql/issues/543#issuecomment-777069712)

My bad I hadn't seen that comment :/

glensc:
@tianon: can you have look (and mark the existing comments as out of scope)?

yosifkit:
Yeah, time zones are not available until after initialization, so this makes sense to add. CC @tianon & @ltangvald

I think alternatively they can set `TZ`, but it makes sense to support it in config files and flags too.

tianon:
I don't love it, but it makes sense. :+1:

PR #741, opened 2021-02-16
Add consistent variable for "--verbose --help" arguments

tianon:
Given these are duplicated (or should be -- as seen in the diff we already messed this up!) it makes sense to make them share a common variable.

This is ported from the changes I've proposed in https://github.com/docker-library/mariadb/pull/340 (where MariaDB needs an additional parameter).

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -78,8 +78,14 @@ docker_process_init_files() {
 	done
 }
 
+# arguments necessary to run "mysqld --verbose --help" successfully (used for testing configuration validity and for extracting default/configured values)
+_verboseHelpArgs=(
+	--verbose --help
+	--log-bin-index="$(mktemp -u)" # https://github.com/docker-library/mysql/issues/136
+)
+
 mysql_check_config() {
-	local toRun=( "$@" --verbose --help ) errors
+	local toRun=( "$@" "${_verboseHelpArgs[@]}" ) errors
 	if ! errors="$("${toRun[@]}" 2>&1 >/dev/null)"; then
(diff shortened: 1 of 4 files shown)


yosifkit:
For my (future) reference on why `--encrypt-tmp-files`  from https://github.com/docker-library/mariadb/pull/340 was not added here; the flag is not available in MySQL.

```console
$ docker run -it --rm mysql:5.6 --verbose --help | grep encrypt
  --block-encryption-mode=name 
  --des-key-file=name Load keys for des_encrypt() and des_encrypt from given
block-encryption-mode                                      aes-128-ecb
$ docker run -it --rm mysql:5.7 --verbose --help | grep encrypt
  --block-encryption-mode=name 
  --des-key-file=name Load keys for des_encrypt() and des_encrypt from given
block-encryption-mode                                        aes-128-ecb
$ docker run -it --rm mysql:8.0 --verbose --help | grep encrypt
  --binlog-encryption Enable/disable binary and relay logs encryption.
  --binlog-rotate-encryption-master-key-at-startup 
                      Force binlog encryption master key rotation at startup
  --block-encryption-mode=name 
  --default-table-encryption 
                      encryption property unless the user specifies an explicit
                      encryption property.
  --innodb-redo-log-encrypt 
  --innodb-undo-log-encrypt 
  --table-encryption-privilege-check 
binlog-encryption                                            FALSE
binlog-rotate-encryption-master-key-at-startup               FALSE
block-encryption-mode                                        aes-128-ecb
default-table-encryption                                     FALSE
innodb-redo-log-encrypt                                      FALSE
innodb-undo-log-encrypt                                      FALSE
table-encryption-privilege-check                             FALSE
```

Issue #750, opened 2021-03-18
root user can't create new database


Sixeight:
# Problem

- Following pull-request makes prevent creating new database by root user.
  -  https://github.com/docker-library/mysql/pull/749
- Use this container on CI, I want to create database at the launching time.
  - example: use [circleci/mysql:5.7-ram](https://hub.docker.com/layers/circleci/mysql/5.7-ram/images/sha256-e07dd453adeae4a4d87f6a151710950541601d3188adb9405e4085ecaa4909ae?context=explore) , I can't create database before luanch.
- I think following code are wrong condition or wrong error message.
  - https://github.com/docker-library/mysql/blob/b7e01968f8f78ad8e4786be59eeec231541e9c58/5.7/docker-entrypoint.sh#L156-L165
- Message order to set environment  variable, but all environment are ignored by this condition.

# Expected result
- I want to use root user for CI testing.

davidstosik:
We started facing the same issue when CircleCI switched our mysql image to 5.6.51 (previously was 5.6.50).

Sixeight:
@davidstosik You can avoid the error, use `circleci/mysql:5.7.32-ram` to use old `docker-entrypoint.sh` .

davidstosik:
@Sixeight Yeah, I'm about to freeze the mysql image to 5.6.50. 👍🏻

polamjag:
We encountered probably same problem (in image labeled `mysql:5.6`) and this mitigation worked:  https://github.com/docker-library/mysql/issues/750#issuecomment-801565537

ogii:
On CircleCI, It is also possible to specify the image using a previous working SHA

```
- image: circleci/mysql@sha256:eb6d35250a42f16994dea0726d62453c11c956e8a7f5f4fc946c76041bb2ab86
```

where the SHA can be found by checking the logs under the container step log

Sixeight:
Also avoid this issue, Use another user name If you can .

```shell-session
MYSQL_USER=user
MYSQL_PASSWORD=pass
```

davidstosik:
> Also avoid this issue, Use another user name If you can .
> 
> ```shell
> MYSQL_USER=user
> MYSQL_PASSWORD=pass
> ```

A note on this: that work-around indeed works, but does not prevent the `root` user to be created with no password, and then used by our app. [See the docs.](https://hub.docker.com/_/mysql/)

[...]

PR #751, opened 2021-03-18
Correct the error condition introduced by #749

yasaichi:
This PR corrects the error condition introduced by #749.
The condition is wrong so that a root user can't create a new database even when the password is given by `MYSQL_ROOT_PASSWORD` (cf. #750).


yasaichi:
Sorry, I found out that this issue should be resolved by circleci/mysql (e.g. https://github.com/circleci/circleci-images/pull/540). 
Closing.

PR #773, opened 2021-05-20
Add .sql.zst support to docker-entrypoint-initdb.d

hectorm:
Zstandard outperforms gzip and xz in many metrics, it would be beneficial to support it.

Mirroring additions done in MariaDB (MariaDB/mariadb-docker#376) and PostgreSQL (docker-library/postgres#846).


yosifkit:
Sorry for the delay and thanks for the PR. I've done a rebase, so once the tests finish, I think it should be good to go.

PR #782, opened 2021-08-16
Treat an existing but empty /var/lib/mysql as nonexistent database

dallaylaen:
I'm trying to setup mysql in a local minikube environment and I want the storage to be persistent across pod restarts. So I have to mount `/var/lib/mysql`, but then mysql refuses to execute scripts in `/docker-entrypoint-initdb.d`. So I added a second check to make sure that the data directory is not only existing but also nonempty.

I can make a workaround (like, log into the container and run the scripts!) but I would prefer to have a generic solution.

See also https://stackoverflow.com/questions/38504257

### DIFF
--- a/8.0/docker-entrypoint.sh
+++ b/8.0/docker-entrypoint.sh
@@ -220,8 +220,10 @@ docker_setup_env() {
 	file_env 'MYSQL_PASSWORD'
 	file_env 'MYSQL_ROOT_PASSWORD'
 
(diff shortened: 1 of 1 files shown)


dallaylaen:
It looks like the mysql server itself doesn't like an "existing but empty" directory. Will re-check and reopen if I get this to work. Sorry for premature PR...

dallaylaen:
Retested this again, this PR totally misses the point. The entrypoint script is already working as it should.

Issue #562, opened 2019-05-07
Missing directories for a default installation


cezmunsta:
A standard installation of MySQL (both for Oracle MySQL and Percona Server for MySQL) will create the following directories:
- `/var/lib/mysql`
- `/var/lib/mysql-files`; default directory used for `secure_file_priv` is `/var/lib/mysql-files`, whereas this is disabled in this image for some reason.
- `/var/lib/mysql-keyring`; default `keyring_file_data` is `/var/lib/mysql-keyring/keyring`, which means that enabling the plugin immediately causes a warning and forces setting the keyring to be part of the MySQL datadir or similar

Additionally, the permissions for the datadir are inconsistent. Here are the default permissions for the installation:
```
=> docker exec oracle-mysql-8.0 ls -ld /var/lib/mysql{,-files,-keyring}
drwxr-x--- 1 mysql mysql 604 May  7 11:53 /var/lib/mysql
drwxr-x--- 1 mysql mysql   0 May  7 11:53 /var/lib/mysql-files
drwxr-x--- 1 mysql mysql   0 Jan 25 23:03 /var/lib/mysql-keyring
```
Here is the permission set for this image:
```
=> docker exec docker-mysql-8.0 ls -ld /var/lib/mysql
drwxr-xr-x 1 mysql mysql 522 May  7 12:09 /var/lib/mysql
```

ltangvald:
Generally, these paths (datadir, keyring and secure-file-priv) need to be mounted outside the image to be useful (datadir is mounted by default as an anonymous volume) for anything besides simple testing. It's generally a bad idea to store state in the container, because it is ephemeral by nature. So having them created by default probably isn't very useful.

It might be a good idea to add some permission handling for these paths if they exist, since a common issue for Docker is that mounted paths don't have the correct permissions to be used from inside the container. e.g. the entrypoint script tries to ensure the mysql user owns datadir

cezmunsta:
@ltangvald if you are suggesting that this is not an issue then I disagree - the `datadir` is created and managed, so do the same for the other default directories. If a user wants to test MySQL then they are forced to create these directories rather than just use them ephemerally with the relevant config options.

tianon:
I don't think he's suggesting it's not an issue -- I think he's suggesting that it *could* be an issue depending on how folks are using it and how these directories get passed in from the host, so it might in fact make sense for us to do something with their permissions during container startup like we do for `/var/lib/mysql`.

tianon:
[...]

PR #797, opened 2021-11-09
Ensure that /var/run/mysqld directory exists

giggsoff:
We create directory in /var/run during image creation, but for example containerd [mount](https://github.com/lf-edge/eve/blob/master/pkg/pillar/vendor/github.com/containerd/containerd/oci/spec.go#L201-L206) tmpfs on /run directory. We can lose content of directory in this case, so we must ensure that it exists.

Signed-off-by: Petr Fedchenkov <giggsoff@gmail.com>

fixes https://github.com/docker-library/mysql/issues/562

### DIFF
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -191,13 +191,44 @@ docker_verify_minimum_env() {
 docker_create_db_directories() {
 	local user; user="$(id -u)"
 
-	# TODO other directories that are used by default? like /var/lib/mysql-files
-	# see https://github.com/docker-library/mysql/issues/562
-	mkdir -p "$DATADIR"
+	local -A dirs=( ["$DATADIR"]=1 )
+	local dir
+	dir="$(dirname "$SOCKET")"
+	dirs["$dir"]=1
+
+	# "datadir" and "socket" are already handled above (since they were already queried previously)
+	local conf
+	for conf in \
+		general-log-file \
+		keyring_file_data \
+		pid-file \
+		secure-file-priv \
+		slow-query-log-file \
+	; do
+		dir="$(mysql_get_config "$conf" "$@")"
+
+		# skip empty values
+		if [ -z "$dir" ]; then
+			continue
+		fi
+		case "$conf" in
+			secure-file-priv)
+				# already points at a directory
+				;;
+			*)
+				# other config options point at a file, but we need the directory
+				dir="$(dirname "$dir")"
+				;;
+		esac
+
+		dirs["$dir"]=1
+	done
+
+	mkdir -p "${!dirs[@]}"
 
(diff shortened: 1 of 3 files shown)


tianon:
Thanks!

I think instead of hard-coding `/var/run/mysqld`, we should probably check the relevant configuration flag values (like we do for `datadir`).  In this case, I think that's going to be the directory portion of both `pid-file` and `socket`.

I also think we should include this as part of the existing `docker_create_db_directories` function.  It's a bit verbose, but maybe something like this:

```bash
# creates folders for the database
# also ensures permission for user mysql if run as root
docker_create_db_directories() {
	local user; user="$(id -u)"

	local -A dirs=( ["$DATADIR"]=1 )
	local dir
	dir="$(dirname "$SOCKET")"
	dirs["$dir"]=1

	# TODO /var/lib/mysql-files and /var/lib/mysql-keyring ?
	# see https://github.com/docker-library/mysql/issues/562

	# "datadir" and "socket" are already handled above (since they were already queried previously)
	local conf
	for conf in \
		general-log-file \
		pid-file \
		slow-query-log-file \
	; do
		dir="$(mysql_get_config "$conf" "$@")"
		dir="$(dirname "$dir")" # each of these config options points at a file
		dirs["$dir"]=1
	done

	mkdir -p "${!dirs[@]}"

	if [ "$user" = "0" ]; then
		# this will cause less disk access than `chown -R`
		find "${!dirs[@]}" \! -user mysql -exec chown mysql '{}' +
	fi
}
```

Which also means we need to change the usage further down to include `"$@"`:

```bash
		mysql_check_config "$@"
		# Load various environment variables
		docker_setup_env "$@"
		docker_create_db_directories "$@"
```

Please let me know if you'd like to continue working on this or if you'd rather have us take over. :+1:

giggsoff:
Thank you for your comments! It looks reasonable to not hardcode paths, so, I use your suggested code and also add check for absolute path received from dirname (to not handle "." from dirname command).

yosifkit:
Rebased and added in `secure_file_priv` too.

cc @tianon and @ltangvald Does this seem good to go? Should we also add `keyring-file-data` (default: `/var/lib/mysql-keyring/keyring`), so it is also accounted for? It only appears in the help output if the plugin is loaded with `--early-plugin-load=keyring_file.so`.

tianon:
Oh, nice find on `keyring-file-data` - IMO we should add it, yeah (and probably use `case` instead of `if !=` so it's easier to maintain if the list grows).

yosifkit:
Added `keyring-file-data` and switched to a `case` statement.

tianon (approved):

tianon:
@ltangvald I'm going to merge this, but please feel free to let us know if you don't like it and we can revert / continue discussing :heart:

PR #798, opened 2021-11-14
update GOSU to 1.14

ducksecops:
Signed-off-by: daniel sutton <daniel@ducksecops.uk>

Upgradet GOSU 1.14 from prevous GOSU 1.12. Newer version is based on GO 1.16.7 which is supported.


tianon:
I don't think there's anything _super_ compelling in the set of changes between `gosu` 1.12 and 1.14, but it seems reasonable to upgrade anyhow.

PR #805, opened 2022-01-11
refactor: recurse docker-entrypoint-initdb.d dir files

vast0906:
can use docker-entrypoint-initdb.d recurse dir files to import

### DIFF
--- a/5.6/docker-entrypoint.sh
+++ b/5.6/docker-entrypoint.sh
@@ -393,7 +393,7 @@ _main() {
 			mysql_note "Temporary server started."
 
 			docker_setup_db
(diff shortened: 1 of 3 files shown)


yosifkit:
Closing since this is not something we want to add. 

Since initdb scripts can be sourced (https://github.com/docker-library/mysql/pull/656) and everything in the script is functionalized (https://github.com/docker-library/mysql/pull/471), a custom script in initdb can use the `docker_process_init_files` function to recurse as deeply as desired.

Related issues https://github.com/docker-library/postgres/issues/179, https://github.com/docker-library/postgres/issues/605, https://github.com/docker-library/mysql/issues/192, https://github.com/docker-library/mysql/issues/193

PR #808, opened 2022-01-17
Update to new MySQL GPG Key

ltangvald:
As part of the 8.0.28 release, MySQL is introducing a new gpg key. It's a new key instead of extending the duration of the old one because the encryption algorithm has been changed.


ltangvald:
The build failure is expected, and should (fingers-crossed) be resolved when the next MySQL release is out.

ltangvald:
One complication I didn't consider is that we no longer update 5.6, so that signature is old. We could make it so 5.6 uses the old key, but that key expires soon and will not be updated (because 5.6 is no longer supported). I would suggest removing 5.6 builds.

rfay:
What I've been doing with unsupported upstreams is just to remove the key and the /etc/apt/sources.list.d entry from them after build. Then the key can't cause trouble, but other things in the image may still be updatable by child images.

tianon:
Ahhh, I've looked and looked for a canonical place that says "5.6 is EOL" and all I can find is wishy-washy "EOL on X" notices, presumably because it's still technically "supported" but maybe only if you have an active support contract or something?  If the public/open version is EOL, we should definitely just remove it from the repository entirely. :sweat_smile:

tianon:
@ltangvald I've cherry-picked #810 into this one so we can kill both birds here, but I'm happy to pull that back out if you prefer? :heart:

rfay:
Please don't remove 5.5 or 5.6 from dockerhub, as ddev uses those for upstream. They're only used by people who are resurrecting and updating old sites, and only used for local dev.

tianon:
Removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent official images from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

ltangvald:
Since the release is out and all looks good, I'll merge this now. As mentioned, removing 5.6 from the code just means we stop building it. The image will still be there (it does mean you'll need to keep the base os layer updated yourself).

JustinAvaLabs:
> Since the release is out and all looks good, I'll merge this now. As mentioned, removing 5.6 from the code just means we stop building it. The image will still be there (it does mean you'll need to keep the base os layer updated yourself).

You broke the GPG key and now the official docker images don't work.
https://github.com/docker-library/mysql/issues/811

ltangvald:
> > Since the release is out and all looks good, I'll merge this now. As mentioned, removing 5.6 from the code just means we stop building it. The image will still be there (it does mean you'll need to keep the base os layer updated yourself).
> 
> You broke the GPG key and now the official docker images don't work. #811

Not actually related to this PR. The official mysql apt repo is using a new signing key. The old one uses DSA1024, which is no longer considered secure.

JustinAvaLabs:
> 

Got it, my mistake.  Is there any insight on how long until the official images are fixed?

ltangvald:
I'm not clear on when/how rebuilds happen for official images. @tianon ?

yosifkit:
Short answer is they are built once the commits here are referenced in the `library/mysql` file in https://github.com/docker-library/official-images. So after https://github.com/docker-library/official-images/pull/11700 is merged.

More info: https://github.com/docker-library/mysql/tree/f4032a1af40618ab81ecfe03ce0366aeb9f5af5e#see-a-change-merged-here-that-doesnt-show-up-on-docker-hub-yet

PR #810, opened 2022-01-18
Remove EOL 5.6

tianon:
Refs https://github.com/docker-library/mysql/pull/808#issuecomment-1015375115

cc @ltangvald


tianon:
Merged as part of #808 :+1:

PR #828, opened 2022-02-22
Switch from "pwgen" to "openssl rand" and add missing Oracle deps

### LINKED ISSUE #826: gunzip missing from mysql:8.0.28-oracle
Entrypoint handles gzipped SQL files by [calling gunzip](https://github.com/docker-library/mysql/blob/6a2fb1deb604a67dd87c0597283ef17289bc9e43/8.0/docker-entrypoint.sh#L76) on them. But when I tried building a container `FROM mysql:8.0.28-oracle` with a gzipped source mounted into `/docker-entrypoint-initdb.d`, I got this:

```
sd_dev_mysql  | 2022-02-19 01:40:42+00:00 [Note] [Entrypoint]: /usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/sd-prod-mysqldump-2022-02-18-08-01-02-UTC.sql.gz
sd_dev_mysql  | /usr/local/bin/docker-entrypoint.sh: line 76: gunzip: command not found
sd_dev_mysql exited with code 127
```

tianon:
Fixes #826

### DIFF
--- a/template/docker-entrypoint.sh
+++ b/template/docker-entrypoint.sh
@@ -245,7 +245,7 @@ docker_setup_db() {
 	fi
 	# Generate random root password
 	if [ -n "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
-		export MYSQL_ROOT_PASSWORD="$(pwgen -1 32)"
+		MYSQL_ROOT_PASSWORD="$(openssl rand -base64 24)"; export MYSQL_ROOT_PASSWORD
 		mysql_note "GENERATED ROOT PASSWORD: $MYSQL_ROOT_PASSWORD"
 	fi
 	# Sets root password and creates root users for non-localhost hosts
--- a/5.7/Dockerfile.debian
+++ b/5.7/Dockerfile.debian
@@ -36,10 +36,9 @@ RUN set -eux; \
 
 RUN mkdir /docker-entrypoint-initdb.d
 
-RUN apt-get update && apt-get install -y --no-install-recommends \
-# for MYSQL_RANDOM_ROOT_PASSWORD
-		pwgen \
-# for mysql_ssl_rsa_setup
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y --no-install-recommends \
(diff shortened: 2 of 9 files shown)


yosifkit (approved):

template/docker-entrypoint.sh
```diff
@@ -245,7 +245,7 @@ docker_setup_db() {
 	fi
 	# Generate random root password
 	if [ -n "$MYSQL_RANDOM_ROOT_PASSWORD" ]; then
-		export MYSQL_ROOT_PASSWORD="$(pwgen -1 32)"
+		MYSQL_ROOT_PASSWORD="$(openssl rand -hex 16)"; export MYSQL_ROOT_PASSWORD
```
tianon, line 248: This is _technically_ a pretty major decrease in bits of entropy, but I'm not sure whether it's important.

The current `pwgen` is very roughly somewhere around (26×2+10)^32 = ~2.3×10^57, where this is 8^16 = ~2.8×10^14.

This is essentially the same as MD5 -- if we're worried about it, we could double it to 32 and get to the same length as a SHA256, but I'm not sure it makes much difference? (also want to be wary of MySQL's password length limits, which I seem to recall being something close to 32 characters?)

We could also increase our entry via something like base64'd random data instead (and trim it to the appropriate length), ala something like `openssl rand -base64 32 | cut -b -32`.

ltangvald: No. I think those symbols should be fine (might actually be a requirement to have some special symbols if users enable the password_validation plugin).

ltangvald: openssl rand -base64 isn't guaranteed to include special characters required by the password_validation plugin. However, that plugin isn't enabled by default, and pwgen as used in the Debian-based images never includes special characters, so I would suggest at most making a little note about it in the README.

ltangvald (approved):

tianon:
Thanks for reviewing and adding your thoughts, @ltangvald! :heart:

PR #830, opened 2022-02-24
Update socket location for mysql clients

### LINKED ISSUE #829: Wrong socket path when connecting from within container
Steps to reproduce:
1. Start a container `docker run --name mysql-test -e MYSQL_ROOT_PASSWORD=password -d mysql:8.0-oracle`
2. Open a shell inside container `docker exec -it mysql-test bash`
3. Execute `mysql` or `mysql -u root -p` or `mysql -u root -ppassword`

Throws error `ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/lib/mysql/mysql.sock' (2)`.

Real socket path is `/var/run/mysqld/mysqld.sock`, so I really don't know how it tries to connect to `/var/lib/mysql/mysql.sock`.

PS: I know I can specify the host, as `mysql -h 127.0.0.1 -u root -p`, and it works.

yosifkit:
The updated `my.cnf` is enough for `mysql` but `mysqlsh --mysql` still uses the compiled in value, so we need the symlink in the data directory. `mysqlsh [user]@localhost` works since it uses the mysqlx port of `33060` by default.

Fixes #829

### DIFF
--- a/template/docker-entrypoint.sh
+++ b/template/docker-entrypoint.sh
@@ -104,6 +104,18 @@ mysql_get_config() {
 	# match "datadir      /some/path with/spaces in/it here" but not "--xyz=abc\n     datadir (xyz)"
 }
 
+# Ensure that the package default socket can also be used
+# since rpm packages are compiled with a different socket location
+# and "mysqlsh --mysql" doesn't read the [client] config
+# related to https://github.com/docker-library/mysql/issues/829
+mysql_socket_fix() {
+	local defaultSocket
+	defaultSocket="$(mysql_get_config 'socket' mysqld --no-defaults)"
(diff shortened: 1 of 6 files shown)


template/docker-entrypoint.sh
```diff
@@ -379,6 +391,7 @@ _main() {
 			mysql_note "MySQL init process done. Ready for start up."
 			echo
 		fi
+		mysql_socket_fix
```
tianon, line 394: For `mysqlsh` to work in `docker-entrypoint-initdb.d` scripts too, this needs to happen after `docker_init_database_dir` too, right? (maybe this becomes `else` instead of outside the `if`?)

tianon (approved):

ltangvald (approved):
LGTM!

PR #832, opened 2022-02-25
add missing openssl dependency

porkotron:
Fixes [#831](https://github.com/docker-library/mysql/issues/831)

### DIFF
--- a/5.7/Dockerfile.oracle
+++ b/5.7/Dockerfile.oracle
@@ -41,6 +41,7 @@ RUN set -eux; \
 RUN set -eux; \
 	yum install -y \
 		gzip \
+		openssl \
 		xz \
 	; \
 	yum clean all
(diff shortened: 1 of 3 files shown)


tianon:
This needs to be added to https://github.com/docker-library/mysql/blob/0c0f7399dcc4a8e97f41f4bf558359d5bed541d5/template/Dockerfile.oracle#L36-L45 too (and should be alphabetical):

```dockerfile
RUN set -eux; \
	{{ dnf }} install -y \
		gzip \
		openssl \
		xz \
{{ if .oracle.variant | startswith("7") then "" else ( -}}
# Oracle Linux 8+ is very slim :)
		findutils \
{{ ) end -}}
	; \
	{{ dnf }} clean all
```

(I'm also happy to take over from here if you'd rather.  Thanks for catching/reporting the issue! :+1:)

porkotron:
Fixed that but yea you can take it from here

PR #866, opened 2022-06-13
Update 8.0-debian to bullseye

tianon:


tianon:
IMO we should move forward with this on Tuesday of next week right after the US holiday.

PR #867, opened 2022-06-13
Update "oracle" variants to be the default variants

tianon:
The Oracle Linux-based images:

- are smaller (by a pretty non-trivial amount - nearly 100MiB)
- support more architectures (arm64!)
- are more supported/recommended by MySQL upstream

I do not believe it is common for users to install additional packages in images that are `FROM mysql`, but if it is, the Debian-based images are not going away with this change (those users will just have to adapt - either by switching their package installations to Oracle-based packages or by using the Debian-based variants explicitly).

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -8,8 +8,8 @@ declare -A aliases=(
 
 defaultDefaultVariant='oracle'
 declare -A defaultVariants=(
-	[5.7]='debian'
-	[8.0]='debian'
+	#[5.7]='debian'
+	#[8.0]='debian'
 )
 
 self="$(basename "$BASH_SOURCE")"


tianon:
LOL, I just started making this exact same PR a second time:

> Update "oracle" variants to be the default variants
> 
> To reiterate the rationale:
> - more "upstream-supported" configuration
> - more architectures (hello, `arm64v8`!)
> - smaller images 👀
> - `mysqlsh`

These variants have been available since February (#680), and we're not planning to remove the Debian-based variants of 5.7 or 8.0 anytime soon (however, if there's ever an 8.1 or 9.0, that will likely be Oracle-only).

IMO we should move forward with this on Tuesday of next week right after the US holiday.

yosifkit (approved):

PR #851, opened 2022-05-18
Added support for bunzip2 compressed DB dumps

woutersamaey:

PR #870, opened 2022-06-24
Added "bzip2" so we can import .sql.bz2 dumps.

woutersamaey:
Follow-up to #851

### DIFF
--- a/5.7/Dockerfile.debian
+++ b/5.7/Dockerfile.debian
@@ -39,6 +39,7 @@ RUN mkdir /docker-entrypoint-initdb.d
 RUN set -eux; \
 	apt-get update; \
 	apt-get install -y --no-install-recommends \
+		bzip2 \
 		openssl \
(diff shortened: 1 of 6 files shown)


tianon:
Nice catch!  Can we move this down to be in the same block of package installs as the `bzip2` cohorts, `zstd` and `xz-utils`? :innocent:

(The lists of packages should also be sorted alphabetically, so this will go at the top of the block of packages lower down.)

If you'd rather, I'd also be happy to take over this change from here. :+1:

woutersamaey:
I can do it on monday (if my day isn’t too hectic), but if you can do it faster then by all means, go ahead!

woutersamaey:
@tianon I moved the package further down the file and should be OK now?

tianon:
Force pushed to fix whitespace and update Oracle variants too :+1: :heart:

woutersamaey:
This PR is merged and closed, but I'm still seeing issues with tag `mysql:5.7-debian`. Is Dockerhub not updated somehow?
Or is there a delay?
How can I make sure I got the latest version of `mysql:5.7-debian`

PR #885, opened 2022-07-08
Add "/entrypoint.sh" backwards-compat symlink in Oracle-based images

### LINKED ISSUE #884: Dockerfile difference between revision releases
The debian->oracle image switch omits the `/entrypoint.sh` symlink

The applicable line from a debian Dockerfile is the following:
https://github.com/docker-library/mysql/blob/master/5.7/Dockerfile.debian#L96

While that symlink is itself a backwards compat thing, I don't think its the sort of thing which should be removed from what is otherwise just a revision update from `5.7.37` to `5.7.38` (for tags like `:latest` I agree that all bets are off, but I think there should be a degree of consistency between releases in the same major.minor series)

If you were looking to keep the oracle Dockerfiles clean but still have those be the new foundation, perhaps until a new release series comes to exist, the former debian defaulted tags such as `:5.7` could build `FROM mysql:5.7-oracle`, simply adding things like the entrypoint symlink, a symlink for the config dir changes which were mentioned in other issues. It might also allow a more explicit means to warn of deprecated changes in that image series

tianon:
Fixes #884

Thanks for the report, @martynd! :heart:

### DIFF
--- a/template/Dockerfile.oracle
+++ b/template/Dockerfile.oracle
@@ -130,6 +130,9 @@ RUN set -eux; \
 VOLUME /var/lib/mysql
 
 COPY docker-entrypoint.sh /usr/local/bin/
+{{ if [ "5.7", "8.0" ] | index(env.version) then ( -}}
+RUN ln -s usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
+{{ ) else "" end -}}
 ENTRYPOINT ["docker-entrypoint.sh"]
 
 EXPOSE 3306 33060
--- a/5.7/Dockerfile.oracle
+++ b/5.7/Dockerfile.oracle
@@ -115,6 +115,7 @@ RUN set -eux; \
 VOLUME /var/lib/mysql
 
 COPY docker-entrypoint.sh /usr/local/bin/
+RUN ln -s usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
 ENTRYPOINT ["docker-entrypoint.sh"]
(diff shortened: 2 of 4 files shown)


template/Dockerfile.oracle
```diff
@@ -130,6 +130,9 @@ RUN set -eux; \
 VOLUME /var/lib/mysql
 
 COPY docker-entrypoint.sh /usr/local/bin/
+{{ if [ "5.7", "8.0" ] | index(env.version) then ( -}}
+RUN ln -s usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
```
thaJeztah, line 134: Should this use an absolute path? (just to be sure / explicit)?
```suggestion
RUN ln -s /usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
```

thaJeztah: (I guess the image would unlikely get a WORKDIR changed/set, but if it would..)

tianon: :shrug: it's put in the root, and this matches the Debian version :innocent:

tianon: (Semi-intentional that it's not "pretty" so that we definitely get rid of it later :eyes:)

thaJeztah: 🤷 alright, you win 😂💜

thaJeztah: No hard(link) feelings 🤣

Issue #887, opened 2022-07-10
can't create lock file /var/run/mysqld/mysqlx.sock.lock


ptr727:
I use watchtower to keep containers up to date, and I use the `latest` tag.  
I just noticed Home Assistant reporting `History integration disabled`.

Troubleshooting shows that MySQL is no longer working:

```
2022-07-10 02:13:29+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.29-1.el8 started.
'/var/lib/mysql/mysql.sock' -> '/var/run/mysqld/mysqld.sock'
2022-07-10T02:13:30.216474Z 0 [System] [MY-010116] [Server] /usr/sbin/mysqld (mysqld 8.0.29) starting as process 1
2022-07-10T02:13:30.219843Z 0 [Warning] [MY-010122] [Server] One can only use the --user switch if running as root
2022-07-10T02:13:30.230068Z 1 [System] [MY-013576] [InnoDB] InnoDB initialization has started.
2022-07-10T02:13:30.421841Z 1 [System] [MY-013577] [InnoDB] InnoDB initialization has ended.
2022-07-10T02:13:30.559703Z 0 [ERROR] [MY-011300] [Server] Plugin mysqlx reported: 'Setup of socket: '/var/run/mysqld/mysqlx.sock' failed, can't create lock file /var/run/mysqld/mysqlx.sock.lock'
2022-07-10T02:13:30.672953Z 0 [Warning] [MY-010068] [Server] CA certificate ca.pem is self signed.
2022-07-10T02:13:30.673008Z 0 [System] [MY-013602] [Server] Channel mysql_main configured to support TLS. Encrypted connections are now supported for this channel.
2022-07-10T02:13:30.673560Z 0 [ERROR] [MY-010273] [Server] Could not create unix socket lock file /var/run/mysqld/mysqld.sock.lock.
2022-07-10T02:13:30.673593Z 0 [ERROR] [MY-010268] [Server] Unable to setup unix socket lock file.
2022-07-10T02:13:30.673909Z 0 [ERROR] [MY-010119] [Server] Aborting
2022-07-10T02:13:32.078735Z 0 [System] [MY-010910] [Server] /usr/sbin/mysqld: Shutdown complete (mysqld 8.0.29)  MySQL Community Server - GPL.
```

[...]

Issue #880, opened 2022-07-07
Loglevel behaviour of mysql image changed


DrVanScott:
Hi,

probably since https://github.com/docker-library/mysql/pull/867 the behaviour of the mysql daemon changed slightly. Before of that chance the log of the container also contained messages of the "[NOTE]" and "[WARNING]" loglevel. This can now be achieved again by explicitly refer to mysql:\*-debian instead of just mysql:\*

I think there shouldn't be such a difference between these variants. Is it possible to include both loglevels also in the oracle variant?

tianon:
Oh interesting -- back in https://github.com/docker-library/mysql/pull/680 I was doing detailed comparisons of the full `mysqld` configuration (see https://github.com/docker-library/mysql/pull/680#issuecomment-1040584661 for the latest one I did).

Any idea which configuration option this would be so we can dig more effectively into why they're different?

DrVanScott:
Unfortunately not, but maybe this issue (i suspect it to be a bug, btw) is already fixed by https://github.com/docker-library/mysql/pull/885 ?

PR #889, opened 2022-07-12
Fix arbitrary user support on oracle-linux based images

yosifkit:
The rpm package has different opinions on the mode of `/var/run/mysqld`, so we have to adjust permissions after install.

When testing running as an arbitrary user, I also found that the `my.cnf` in 5.7 has a `log-error=/var/log/mysqld.log` line which causes startup to break; so this should hopefully fix the missing log level in #880 since they have likely been going to a file :disappointed: :see_no_evil: 

Fixes https://github.com/docker-library/mysql/issues/887 resolves https://github.com/docker-library/mysql/issues/880

### DIFF
--- a/template/Dockerfile.oracle
+++ b/template/Dockerfile.oracle
@@ -17,14 +17,7 @@ FROM oraclelinux:{{ .oracle.variant }}
 
 RUN set -eux; \
 	groupadd --system --gid 999 mysql; \
-	useradd --system --uid 999 --gid 999 --home-dir /var/lib/mysql --no-create-home mysql; \
-	\
-	mkdir /var/lib/mysql /var/run/mysqld; \
-	chown mysql:mysql /var/lib/mysql /var/run/mysqld; \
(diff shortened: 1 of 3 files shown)


template/Dockerfile.oracle
```diff
@@ -108,7 +101,21 @@ RUN set -eux; \
 # 5.7 Debian-based images also included "/etc/mysql/mysql.conf.d" so let's include it too
 	{ echo '!includedir /etc/mysql/mysql.conf.d/'; } >> /etc/my.cnf; \
 	mkdir -p /etc/mysql/mysql.conf.d; \
+	\
+# comment out a few problematic configuration values
+	find /etc/my.cnf /etc/mysql/ -name '*.cnf' -print0 \
+		| xargs -0 grep -lZE '^(bind-address|log)' \
+		| xargs -rt -0 sed -Ei 's/^(bind-address|log)/#&/'; \
```
yosifkit, line 108: These lines are copied from the Debian Dockerfile even though we don't have a `bind-address` in the `cnf` file.

PR #912, opened 2022-08-24
Ignore "NULL" values in mkdir loop

### LINKED ISSUE #911: Unable to run Container of Image "mysql:8.0.30-debian" as non-root User
#### Issue description
After starting the container with command ``docker run -u 1000:1000 -d mysql:8.0-debian`` it crashes during the startup with the following logs:
```
2022-08-24 13:20:04+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.30-1debian11 started.
2022-08-24T13:20:04.692996091Z mkdir: cannot create directory 'NULL': Permission denied
```
This does not happen if the container is run as root user.

#### What's the expected result?
To my understanding, the container should start up normally, as it did prior to commit https://github.com/docker-library/mysql/commit/f1f8eea87223e7c501c16fab7c4e8e023d42a896.

#### Additional details / screenshot
OS: Ubuntu 22.04.1 LTS
Docker Engine Version: 20.10.17, build 100c701
MySQL Image Version: mysql:8.0.30-debian

tianon:
Fixes #911

### DIFF
--- a/5.7/docker-entrypoint.sh
+++ b/5.7/docker-entrypoint.sh
@@ -208,7 +208,7 @@ docker_create_db_directories() {
 		dir="$(mysql_get_config "$conf" "$@")"
 
 		# skip empty values
-		if [ -z "$dir" ]; then
+		if [ -z "$dir" ] || [ "$dir" = 'NULL' ]; then
 			continue
 		fi
 		case "$conf" in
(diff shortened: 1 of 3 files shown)


tianon:
Tested with @wglambert's example from https://github.com/docker-library/mysql/issues/911#issuecomment-1225935013 successfully :eyes:

PR #923, opened 2022-10-26
Allow for MySQL user w/ empty password

nVitius:
The value of `MYSQL_ALLOW_EMPTY_PASSWORD` currently only affects the root user. This flag should also determine if the `MYSQL_USER` can be created with an empty password.

### DIFF
--- a/8.0/docker-entrypoint.sh
+++ b/8.0/docker-entrypoint.sh
@@ -179,7 +179,7 @@ docker_verify_minimum_env() {
 	fi
 
 	# warn when missing one of MYSQL_USER or MYSQL_PASSWORD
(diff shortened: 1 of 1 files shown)


yosifkit:
`MYSQL_ALLOW_EMPTY_PASSWORD` has been just for the `root` user since it was added in 2014 (https://github.com/docker-library/mysql/pull/15) and is [documented](https://github.com/docker-library/docs/tree/8cd127622773da50ba45cc7af74f40672877b666/mysql#mysql_allow_empty_password) as such:
> Set to a non-empty value, like `yes`, to allow the container to be started with a blank password for the root user.

While it was a bad choice of name in that it doesn't specify `ROOT` in it (just like `MYSQL_ONETIME_PASSWORD`), we have no intention of expanding its functionality at this time.

PR #973, opened 2023-05-18
Added some required mysql packages to install list

adampweb:
Added **some required mysql packages** to install list which are **not installing automatically** of unidentified reason. Without these dependencies the build process fails with "unmet dependencies" error:

The following packages have unmet dependencies:
  mysql-community-client
  Depends: mysql-common (>= 8.0.33-1debian11)
  Depends: mysql-community-client-core (= 8.0.33-1debian11) but it is not installable

**After that need to regenerate the Dockerfile (for Debian)**

### DIFF
--- a/template/Dockerfile.debian
+++ b/template/Dockerfile.debian
@@ -80,6 +80,9 @@ RUN { \
 # don't reverse lookup hostnames, they are usually another container
 	&& echo '[mysqld]\nskip-host-cache\nskip-name-resolve' > /etc/mysql/conf.d/docker.cnf \
 {{ ) else ( -}}
+        mysql-common="${MYSQL_VERSION}" \
+        mysql-community-client-plugins="${MYSQL_VERSION}" \
+        mysql-community-client-core="${MYSQL_VERSION}" \
 		mysql-community-client="${MYSQL_VERSION}" \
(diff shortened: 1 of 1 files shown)


tianon:
Can you elaborate?  Our builds aren't failing (you can see a very very recent successful build with of all our existing `Dockerfile`s as-is at https://github.com/docker-library/mysql/actions/runs/4969497525).

adampweb:
> Can you elaborate? Our builds aren't failing (you can see a very very recent successful build with of all our existing `Dockerfile`s as-is at https://github.com/docker-library/mysql/actions/runs/4969497525).

I copied the files from [here](https://github.com/docker-library/mysql/tree/master/8.0). Only changed the `FROM` line in dockerfile because I used my upgraded Debian image from my private registry ([Dockerfile for my modified image](https://pastebin.com/7LuiK7ga))
After that when I tried to build image I received error message which detailed in above comments. After when I added the 3 new package to your Dockerfile and re-run the build command completed successfully.

yosifkit:
The reason they are not installing automatically is because of your `mysql*` pin priority; you probably need an `origin` in order for that to not affect the packages coming from MySQL apt repos. https://manpages.debian.org/bullseye/apt/apt_preferences.5.en.html

PR #981, opened 2023-07-03
Add "suite aliases" ("-oraclelinux8", "-bullseye", etc)

### LINKED ISSUE #980: Named Debian builds
### Overview
When building a LAMP stack, all of the other component images from docker-library are based on named Debian versions such as `buster`, `bullseye`, and `bookworm`, and can reuse the base OS layer between them. Only MySQL uses generic `-debian` tags, where it's not immediately clear which version is being used. Instead of easily setting all the components to the same base image, I have to manually check which OS version is used by MySQL, and then pick the correct tags for the other components.

Going forward, please consider named version tags for Debian, `:bullseye`, `:bookworm`, etc. So that in the future, I and others like me could simply pull e.g. `mysql:bookworm`, `php:bookworm`, and `httpd:bookworm`, and be sure that they all use the same OS base layer.

### Debian vs Oracle
I saw that you've moved to Oracle Linux by default in #867, so I want to address your reasoning:

> The Oracle Linux-based images:

- "are smaller": This may be true if MySQL is deployed alone, but not when deployed as part of a stack. Reusing the same base OS layer in *all* of the images of the stack ultimately leads to a smaller total footprint than having to use 2 or more OSes because MySQL or some other component decide to be an oddball.
- "support more architectures (arm64!)":  Not true (anymore?). [`debian`](https://hub.docker.com/_/debian/tags?page=1&name=stable) currently supports 8 architectures, including `amd64` and `arm64`. [`oraclelinux`](https://hub.docker.com/_/oraclelinux/tags?page=1) supports only those two.
- "are more supported/recommended by MySQL upstream": Can't really argue with that one, except to repeat that one has to consider MySQL as part of a wider deployment, and not just in isolation. I assume that Debian is still pretty well supported by MySQL.

I'm not asking that you revert your default OS decision, but please reconsider supporting *named* Debian versions in the future.

### Sizes Comparison
I tried the following combinations on Kubuntu 20.04 using Podman 3.4.2 ("Listed Size" from `podman images` output):
```
Image                   Based On                Listed Size     Size On Disk
mysql:latest            oraclelinux:8-slim      579M             574M
php:8.2-fpm-bullseye    debian:11-slim          459M            +451M
httpd:bullseye          debian:11-slim          150M            + 64M

total:                                                          1089M
```
```
Image                   Based On                Listed Size     Size On Disk
mysql:latest            oraclelinux:8-slim      579M             574M
php:8.2-fpm-bullseye    debian:11-slim          459M            +451M
nginx:bullseye          debian:11-slim          147M            + 61M

total:                                                          1086M
```
```
Image                   Based On                Listed Size     Size On Disk
mysql:debian            debian:11-slim          608M             588M
php:8.2-fpm-bullseye    debian:11-slim          459M            +366M
httpd:bullseye          debian:11-slim          150M            + 64M

total:                                                          1018M
```
```
Image                   Based On                Listed Size     Size On Disk
mysql:debian            debian:11-slim          608M             588M
php:8.2-fpm-bullseye    debian:11-slim          459M            +366M
nginx:bullseye          debian:11-slim          147M            + 61M

total:                                                          1015M
```
Deployed alone, `mysql:debian` is indeed a bit larger, but only by 14M actual size on disk. But as part of a stack reusing the same base image, `mysql:debian` is ~70M *smaller* in total disk usage.

Btw, comparing [`mysql:debian`](https://hub.docker.com/layers/library/mysql/debian/images/sha256-8dd2269034dd8680a2d90403977c9604bbaeb0e30364d105340a50241245785c?context=explore) to [`mysql:latest`](https://hub.docker.com/layers/library/mysql/latest/images/sha256-29ea1f451c3aad88df5a24288f76442286dc7fa6874d96a022e93a8a7efda880?context=explore) in Docker Hub, the size difference is actually in the packages and dependencies, not in the base image. The Debian base image is actually smaller.

tianon:
Closes #980

<details>
<summary>Diff:</summary>

```diff
$ diff -u --color <(bashbrew cat mysql) <(bashbrew cat <(./generate-stackbrew-library.sh))
--- /dev/fd/63	2023-07-03 14:47:19.088329349 -0700
+++ /dev/fd/62	2023-07-03 14:47:19.088329349 -0700
@@ -1,23 +1,23 @@
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon), Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/mysql.git
 
-Tags: 8.0.33, 8.0, 8, latest, 8.0.33-oracle, 8.0-oracle, 8-oracle, oracle
+Tags: 8.0.33, 8.0, 8, latest, 8.0.33-oraclelinux8, 8.0-oraclelinux8, 8-oraclelinux8, oraclelinux8, 8.0.33-oracle, 8.0-oracle, 8-oracle, oracle
 Architectures: amd64, arm64v8
 GitCommit: 1bfa4724fe112b4246672ed2b3c42142f17d5636
 Directory: 8.0
 File: Dockerfile.oracle
 
-Tags: 8.0.33-debian, 8.0-debian, 8-debian, debian
+Tags: 8.0.33-bullseye, 8.0-bullseye, 8-bullseye, bullseye, 8.0.33-debian, 8.0-debian, 8-debian, debian
 GitCommit: 611aa464a96f69e5d4d4172b14ca829ded162b42
 Directory: 8.0
 File: Dockerfile.debian
 
-Tags: 5.7.42, 5.7, 5, 5.7.42-oracle, 5.7-oracle, 5-oracle
+Tags: 5.7.42, 5.7, 5, 5.7.42-oraclelinux7, 5.7-oraclelinux7, 5-oraclelinux7, 5.7.42-oracle, 5.7-oracle, 5-oracle
 GitCommit: e4bdff20153e6a05364a8fe80356dcfd502445bd
 Directory: 5.7
 File: Dockerfile.oracle
 
-Tags: 5.7.42-debian, 5.7-debian, 5-debian
+Tags: 5.7.42-buster, 5.7-buster, 5-buster, 5.7.42-debian, 5.7-debian, 5-debian
 GitCommit: 611aa464a96f69e5d4d4172b14ca829ded162b42
 Directory: 5.7
 File: Dockerfile.debian
```

</details>

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -79,17 +79,33 @@ for version; do
 	)
 
 	for variant in oracle debian; do
+		export variant
+
 		df="Dockerfile.$variant"
 		commit="$(dirCommit "$version" "$df")"
 
 		variantAliases=( "${versionAliases[@]/%/-$variant}" )
 		variantAliases=( "${variantAliases[@]//latest-/}" )
+
+		case "$variant" in
+			debian)
+				suite="$(jq -r '.[env.version][env.variant].suite' versions.json)"
+				variantAliases=( "${versionAliases[@]/%/-$suite}" "${variantAliases[@]}" )
+				variantAliases=( "${variantAliases[@]//latest-/}" )
+				;;
+
+			oracle)
+				ol="$(jq -r '.[env.version][env.variant].variant | split("-")[0]' versions.json)"
+				variantAliases=( "${versionAliases[@]/%/-oraclelinux$ol}" "${variantAliases[@]}" )
+				variantAliases=( "${variantAliases[@]//latest-/}" )
+				;;
+		esac
+
 		if [ "$variant" = "$defaultVariant" ]; then
 			variantAliases=( "${versionAliases[@]}" "${variantAliases[@]}" )
 		fi
 
 		# TODO if the list of architectures supported by MySQL ever is greater than that of the base image it's FROM, this list will need to be filtered
-		export variant
 		variantArches="$(jq -r '.[env.version][env.variant].architectures | join(", ")' versions.json)"
 
 		echo


yosifkit:
Should these stick to the "full" Oracle Linux name like the openjdk tags do?

https://github.com/docker-library/official-images/blob/849f600bb19aeeea0bb0f5d022fb85780d9dc7d3/library/openjdk#L7

tianon:
Oh, I thought we'd gone short there, but I can update.

tianon:
PR, title, and OP diff updated

yosifkit (approved):

tianon:
@ltangvald does this seem pretty reasonable?  anything about it you might change? :eyes:

tianon:
I think we should probably merge this :eyes:

PR #1000, opened 2023-10-05
Added attestation for Gosu

LaurentGoderre:


apply-templates.sh
```diff
@@ -13,6 +13,11 @@ elif [ "$BASH_SOURCE" -nt "$jqt" ]; then
 	wget -qO "$jqt" 'https://github.com/docker-library/bashbrew/raw/9f6a35772ac863a0241f147c820354e4008edf38/scripts/jq-template.awk'
 fi
 
+jqf='template-helper-functions.jq'
+if [ "$BASH_SOURCE" -nt "$jqf" ]; then
+	wget -qO "$jqf" 'https://github.com/docker-library/bashbrew/raw/master/scripts/template-helper-functions.jq'
+fi
```
yosifkit, line 19: Can we get the improvements I mentioned on https://github.com/docker-library/redis/pull/379#discussion_r1353673994 over here?

template/Dockerfile.debian
```diff
@@ -1,3 +1,4 @@
+{{ include "template-helper-functions" }}
```
yosifkit, line 1: Need the `-` to tell it to slurp up the whitespace (so that it doesn't generate a blank line in the final Dockerfiles.

```suggestion
{{ include "template-helper-functions" -}}
```

template/Dockerfile.debian
```diff
@@ -26,7 +27,20 @@ RUN set -eux; \
...
+			version: .gosu.version,
+			params: {
+				os_name: "debian",
+				os_version: .debian.suite
+			},
+			licenses: [
+				"Apache-2.0"
+			]
+		} | sbom | tostring
+	}}' > /usr/local/gosu.spdx.json
```
yosifkit, line 43: Rather than just hoping that the quoting will always work correctly, we can just drop the outer quotes and let `jq` escape the string for shell. (We should add this change to all the `sbom` jq invocations that we have added).

```suggestion
	echo {{
		{
			name: "gosu",
			version: .gosu.version,
			params: {
				os_name: "debian",
				os_version: .debian.suite
			},
			licenses: [
				"Apache-2.0"
			]
		} | sbom | tostring | @sh
	}} > /usr/local/gosu.spdx.json
```

Issue #1016, opened 2023-12-14
PGP key expiration


tianon:
Looks like https://bugs.mysql.com/bug.php?id=85029 has resurged today with an expiration of the PGP key used for signing MySQL releases. :sweat_smile:

It appears that RPM/DNF/YUM don't mind the key being expired, but APT sure does, and fails our (re)builds on 8.0 and 5.7 (Debian-based images).

@ltangvald do you think there's a chance of this key getting a renewed expiration date?  If not, we'll probably consider applying something like https://github.com/debuerreotype/debuerreotype/blob/60b625d1ce31bd81525bb67fc3a33f9686bc3433/scripts/.gpgv-ignore-expiration.sh during our build instead (so we still get the cryptographic benefits of PGP but without honoring/failing on the expiration date).

For reference:

```console
root@dddeed483b62:/# wget -qO- 'https://repo.mysql.com/RPM-GPG-KEY-mysql-2022' | gpg --import
gpg: directory '/root/.gnupg' created
gpg: keybox '/root/.gnupg/pubring.kbx' created
gpg: /root/.gnupg/trustdb.gpg: trustdb created
gpg: key 467B942D3A79BD29: public key "MySQL Release Engineering <mysql-build@oss.oracle.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
root@dddeed483b62:/# wget -qO- 'https://repo.mysql.com/RPM-GPG-KEY-mysql' | gpg --import
gpg: key 8C718D3B5072E1F5: public key "MySQL Release Engineering <mysql-build@oss.oracle.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
root@dddeed483b62:/# gpg --fingerprint
/root/.gnupg/pubring.kbx
------------------------
pub   rsa4096 2021-12-14 [SC] [expired: 2023-12-14]
      859B E8D7 C586 F538 430B  19C2 467B 942D 3A79 BD29
uid           [ expired] MySQL Release Engineering <mysql-build@oss.oracle.com>

pub   dsa1024 2003-02-03 [SCA] [expired: 2022-02-16]
      A4A9 4068 76FC BD3C 4567  70C8 8C71 8D3B 5072 E1F5
uid           [ expired] MySQL Release Engineering <mysql-build@oss.oracle.com>

```

disser4:
+1 on this issue!

jinwen-zou-rechargeapps:
+1, breaking many tools.

[...]

PR #1023, opened 2024-01-18
Update innovation release to MySQL 8.3.0.

desrosj:
Happy to make adjustments. I couldn't find good docs on the right steps to take when updating versions, but wanted to make a first pass.

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -25,7 +25,7 @@
   "innovation": {
     "mysql-shell": {
       "repo": "https://repo.mysql.com/yum/mysql-tools-innovation-community/el/8",
-      "version": "8.2.1-1.el8"
+      "version": "8.3.0-1.el8"
     },
     "oracle": {
       "architectures": [
@@ -34,8 +34,8 @@
       ],
       "repo": "https://repo.mysql.com/yum/mysql-innovation-community/docker/el/8",
       "variant": "8-slim",
-      "version": "8.2.0-1.el8"
+      "version": "8.3.0-1.el8"
     },
-    "version": "8.2.0"
+    "version": "8.3.0"
   }
 }


jnoordsij:
Hey there! Thanks for your effort, I also noticed the new image has not appeared yet. These updates should be added automatically, but I noticed the [automatic job triggering this is failing](https://doi-janky.infosiftr.net/job/update.sh/job/mysql/10738/console) due to the new GPG key being used:

> Packaging Notes
Important Change: The GnuPG build key (A8D3785C) used to sign MySQL downloadable packages has been updated. The previous GnuPG build key (3A79BD29) expired on 2023-12-14. For information about verifying the integrity and authenticity of MySQL downloadable packages using GnuPG signature checking, or to obtain a copy of our public GnuPG build key, see [Signature Checking Using GnuPG](https://dev.mysql.com/doc/refman/8.3/en/checking-gpg-signature.html).

Extracted from https://dev.mysql.com/doc/relnotes/mysql/8.3/en/news-8-3-0.html; similarly present on https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-36.html.

To ensure the image builds properly, an update of the `Dockerfile` templates containing updates to the GPG keys is needed. You should be able to add the new key and then run `update.sh` to ensure all templates and versions are properly updated.

If unsure how to proceed let me know, I'd be willing to follow-up with a new MR (or maybe a maintainer is ready to jump in? @tianon 😇).

tianon:
Ouch, yep, thanks @jnoordsij! Looks like #1016 is back and now hitting the RPM repos (as noted by https://github.com/docker-library/mysql/pull/1018 :joy:)

PR #1046, opened 2024-04-30
Add 8.4 (LTS) variant

jnoordsij:
This PR adds 8.4, the new LTS release, as variant. See also https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-0.html.

To do so, the `update.sh` script has been updated to use `"https://repo.mysql.com/yum/mysql-tools-$version-community/el/$oracleVersion"` as new default url for tools, hoping this is the new convention (didn't check for any official statements); see also https://repo.mysql.com/yum/.

No Debian variant was added per https://github.com/docker-library/mysql/pull/867#issuecomment-1171533195.

Moreover, I've taken the liberty to add an `lts` alias, which seems appropriate to me. If not, feel free to (ask me to) remove that.

Finally, when going through the sources I noticed that Oracle Linux 9 has been released a while back. I could not find any support policies regarding OS versions for this image so quickly, but if variants are preferably kept on the same OS for stability, considering to use v9 for this variant is probably appropriate. Otherwise, updating could be left as a separate (unrelated) follow-up PR.

### DIFF
--- a/versions.sh
+++ b/versions.sh
@@ -105,8 +105,8 @@ for version in "${versions[@]}"; do
 
 	rpmRepo="https://repo.mysql.com/yum/mysql-$version-community/docker/el/$oracleVersion"
 	case "$version" in
-		innovation) toolsRepo="https://repo.mysql.com/yum/mysql-tools-innovation-community/el/$oracleVersion" ;;
-		*)          toolsRepo="https://repo.mysql.com/yum/mysql-tools-community/el/$oracleVersion" ;;
+		8.0) toolsRepo="https://repo.mysql.com/yum/mysql-tools-community/el/$oracleVersion" ;;
+		*)   toolsRepo="https://repo.mysql.com/yum/mysql-tools-$version-community/el/$oracleVersion" ;;
 	esac
 	export rpmRepo toolsRepo
 
--- a/versions.sh
+++ b/versions.sh
@@ -172,4 +172,16 @@ for version in "${versions[@]}"; do
 	json="$(jq <<<"$json" -c --argjson doc "$doc" '.[env.version] = $doc')"
 done
 
-jq <<<"$json" -S . > versions.json
+jq <<<"$json" '
+	# sort entries in descending version order so that it is easier to determine later (in "generate-stackbrew-library.sh") if "innovation" should be included or not, and which entry to give the "latest" alias to (the first one!)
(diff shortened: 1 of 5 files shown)


versions.sh
```diff
@@ -105,8 +105,8 @@ for version in "${versions[@]}"; do
 
 	rpmRepo="https://repo.mysql.com/yum/mysql-$version-community/docker/el/$oracleVersion"
 	case "$version" in
-		innovation) toolsRepo="https://repo.mysql.com/yum/mysql-tools-innovation-community/el/$oracleVersion" ;;
```
LaurentGoderre, line 108: Tnere is still an innovation section the the versions files. Shouldn't both lines be added?

yosifkit:
🤔 https://dev.mysql.com/blog-archive/introducing-mysql-innovation-and-long-term-support-lts-versions/
The next `innovation` release will be the first of the `9.x` series and they are roughly quarterly, but not at the same time as an LTS. So, I think we should figure out how to disable `innovation` until we have 9.x (or maybe the `innovation` tags should point at the LTS?)

versions.sh
tianon: In our current code in this section, "innovation" is the special case, but now we've got a new example that matches innovation, so the change here is swapping 8.0 to be the special case going forward instead (now that 5.x is EOL, that seems appropriate). :+1:

yosifkit: For `innovation` that is the value of `$version`, so the new case should work for both.

tianon:
Hmm, yeah -- I think we probably shouldn't tag this new release as `innovation`, since it isn't technically in the innovation "track" (the latest "innovation" release is still 8.3).

I think at most we should update to avoid building `innovation` while any explicit release has a higher version number so that we don't have to revisit this once innovation gets 9.x -- we've got some code for that elsewhere that I could adapt here. :+1:

As usual @jnoordsij, excellent and well-thought-out contribution. :smile:  Thank you! :heart:

tianon:
Oh sorry, I _also_ think the `lts` alias is probably OK?  It's clearly a word upstream is using to describe this. :+1:

tianon:
Oh and further, regarding upgrades like OL9, if upstream supports it (and recommends using it), we're generally in favor of doing it.  Last time I looked, they did not have the Docker-specific package builds we're using available for it, but maybe that's changed, or is different for the new 8.4?

yosifkit (approved):

PR #1047, opened 2024-05-01
Use Oracle Linux 9 as base image

jnoordsij:
This updates the base OS image to Oracle Linux 9, as a follow-up to https://github.com/docker-library/mysql/pull/1046.

However, currently the innovation release only has a Docker-specific package for v8 (see https://repo.mysql.com/yum/mysql-innovation-community/docker/el/), where both 8.0 and 8.4 do have v9 specific releases. Given that the recent 8.0.37 is the only version in the 8.0 series with a v9 package (see https://repo.mysql.com/yum/mysql-8.0-community/docker/el/9/x86_64/), I *suspect* that going forward v9 packages will be released and the next innovation release will also have a v9 variant.

Therefore it's probably wise to:
- Merge this now and make sure to doublecheck this at the next innovation release, to remove the variant override added here;
- Hold off on merging this until the next innovation release (in approx. 3 months) and then update all images at once.

### DIFF
--- a/8.0/Dockerfile.oracle
+++ b/8.0/Dockerfile.oracle
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM oraclelinux:8-slim
+FROM oraclelinux:9-slim
 
 RUN set -eux; \
 	groupadd --system --gid 999 mysql; \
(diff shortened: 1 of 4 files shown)


tianon:
Very nice - I looked at this and was also grappling with "what do we do about innovation?" :smile:

I think this is a good compromise and we can revisit when it gets a new release (since this will work, and give us the new release automatically so that we are likely to notice it, and then it's easy to update).

PR #1076, opened 2024-07-02
Use Oracle Linux 9 as base image for innovation image

jnoordsij:
Follow-up to #1047, this updates the `innovation` image to use Oracle Linux 9 as base OS image. This also allows it to correctly use the latest MySQL release ([9.0.0](https://dev.mysql.com/doc/relnotes/mysql/9.0/en/news-9-0-0.html)) to be used.

For some reason, this prompts the update script to interchange "8.4" and "innovation" in `versions.json`, which combined with updating both the OS and the version to the same number (9) results in a somewhat confusing git diff ;)

### DIFF
--- a/innovation/Dockerfile.oracle
+++ b/innovation/Dockerfile.oracle
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM oraclelinux:8-slim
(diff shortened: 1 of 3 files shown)


tianon:
Aha, the ordering change is intentional to fix "latest" :partying_face:

Very nice, thank you!

PR #1117, opened 2025-07-22
Use Oracle Linux 10 as base image for innovation and 8.4 (lts)

jnoordsij:
Modeled after #1076, this PR updates the base Oracle Linux version to 10 (see also https://blogs.oracle.com/linux/post/oracle-linux-10-now-generally-available and https://github.com/docker-library/official-images/pull/19442).

Note that this is still incomplete, given that while tooling seems available for v10 (see e.g. https://repo.mysql.com/yum/mysql-tools-innovation-community/el/10), the Docker specific builds are not available just yet (see e.g. https://repo.mysql.com/yum/mysql-innovation-community/docker/el/).

Finally there seem to be no v10 builds for 8.0, which makes sense given its expected EOL in April 2026, so this PR for now assumes this will keep v9 as base version.

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -5,13 +5,13 @@
         "amd64",
         "arm64v8"
       ],
-      "repo": "https://repo.mysql.com/yum/mysql-innovation-community/docker/el/9",
-      "version": "9.4.0-1.el9",
+      "repo": "https://repo.mysql.com/yum/mysql-innovation-community/docker/el/10",
+      "version": "9.4.0-1.el10",
       "variant": "9-slim"
     },
     "mysql-shell": {
-      "repo": "https://repo.mysql.com/yum/mysql-tools-innovation-community/el/9",
-      "version": "9.4.0-1.el9"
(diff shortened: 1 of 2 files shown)


yosifkit:
Thanks so much for doing this research. I was wondering if there were OL10 builds just this morning, and you swoop in and answered the question. ❤️

tianon:
Just to have a place to write down that I checked again (with a timestamp), still no MySQL Docker builds (just mysql-tools, and only 8.4 and innovation, as expected). :heart:

**Update (2025-10-21):** still true
**Update (2026-01-21):** still true (and as affirming confirmation, https://github.com/mysql/mysql-docker/commit/65d3a9ad73f0223ffe4a5aadf19db96b4caddffb#diff-90f857fde861cb89017f9c6d7b2281bae25ef8199fc36c9269e7a1ac296a1a52 is also still on OL9)
**Update (2026-07-27):** still true (even post https://blogs.oracle.com/mysql/a-more-predictable-mysql-release-model-calendar-versions-lts-and-innovation)

PR #1133, opened 2026-01-24
Allow to pass mysql uid and gid as build arguments

ninousf:
useful for testing during developpement

### DIFF
--- a/8.0/Dockerfile.oracle
+++ b/8.0/Dockerfile.oracle
@@ -6,9 +6,13 @@
 
 FROM oraclelinux:9-slim
 
+# Args to pass host UID/GID (default to 999 if not provided)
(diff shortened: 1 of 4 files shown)


tianon:
Can you please provide a bit more detail on your use case?  This image supports running as any arbitrary UID/GID, so something like this shouldn't be necessary: https://hub.docker.com/_/mysql#running-as-an-arbitrary-user

ninousf:
It is useful when you want to prepare an image with uid/gid for a specific environment, and in order to avoid the docker user to pass every times --user option

tianon:
So instead of rebuilding this full image, why not do something simpler like this?

```dockerfile
FROM mysql:8.4
USER 1234:5678
```

Issue #1134, opened 2026-02-22
Support for MySQL 8.4 (LTS) in Debian 12 & 13


erikgieseler:
MySQL 8.4 (LTS) packages for Debian 12 and 13 have been available for several months now.
[https://repo.mysql.com/apt/debian/pool/mysql-8.4-lts/m/mysql-community/](https://repo.mysql.com/apt/debian/pool/mysql-8.4-lts/m/mysql-community/)
It would be great to have these in Docker images as an alternative to Oracle Linux.
Thank you!

tianon:
It was an intentional change that 8.0 was the last version to have Debian variants (https://github.com/docker-library/mysql/pull/867), but I do wonder if that's worth revisiting given things like https://github.com/docker-library/mysql/pull/1117 🤔

tianon:
Oh right, we end up having similar issues as #1117 with Debian, as evidenced most obviously by our 8.0 version being stuck on Debian Bookworm (no Trixie packages available).

erikgieseler:
From an admin perspective, going Oracle Linux only is problematic. If the rest of the stack is already Debian-based, introducing a different base OS just for MySQL adds unnecessary divergence. If MySQL 8.4 Debian packages exist, what prevents continuing Debian-based images as well? Is this mainly maintenance or CI overhead? Supporting both variants would provide more flexibility than a single Oracle-only base.

cpt-erwin:
Hi! I took a stab at this since I needed a Debian-based 8.4 image myself. I'm by no means an expert on this project's internals, but after looking at how 8.0 handles Debian support, I noticed that the main difference for 8.4 seems to be that MySQL uses a `-lts` postfix for its APT repository path (i.e., `mysql-8.4-lts` instead of `mysql-8.4`).

Based ont this assumption, I made the following changes in my fork: https://github.com/cpt-erwin/mysql/commit/43d5052ae4b6096603525f5a1d62fcfb811c1280 

**In short:**
- Updated `versions.sh` to handle the `-lts` repo suffix for `8.4`
- Added Debian config to `versions.json` (amd64 / bookworm) by running the script `versions.sh`
- Adjusted Dockerfile.debian template to use mysql-8.4-lts in the apt source line
- Generated 8.4/Dockerfile.debian by running the script `apply-templates.sh`

I tested it locally and it seemed to work ? the container builds and starts up fine with MySQL 8.4.8 on Debian bookworm.

[...]

PR #1140, opened 2026-04-22
Remove 8.0 (EOL)

jnoordsij:
As of 8.0.46 MySQL 8.0 is considered EOL; see also https://dev.mysql.com/doc/relnotes/mysql/8.0/en/.

This PR removes this variant and updates templates accordingly.

It also drops the entire Debian Dockerfile which is now unused. Based on the outcome of #1134, this may still be restored. If it will remain unsupported, an option would be to rename `Dockerfile.oracle` to just `Dockerfile` and remove variant related code.

It is based on top of #1139 and should only be merged in a next iteration, after at least the 8.0.46 image has been built. As the 'general' EOL is "April 2026", it can also be held off slightly longer until the start of May.


yosifkit (approved):
`8.0.46` tags have been published (https://hub.docker.com/_/mysql/tags?name=8.0.46). 🎉 So, LGTM.

PR #1141, opened 2026-05-12
Log which init file failed in docker_process_init_files (fixes #1026)

### LINKED ISSUE #1026: Initialisation script errors aren't logged
I'm not sure if this is intended behaviour but I was running into a problem of not having all my scripts run and wasn't certain why until I tried to execute the sql in the last script it logs that it's initialising and found a bug someone had introduced. Is it not possible for the container to at least show there was an error with the script or better yet log the actual error to track down the error more easily?

pranithreddym:
When a script or SQL file in /docker-entrypoint-initdb.d/ fails during container initialization, the container exits but gives no indication of which file caused the error. This makes debugging difficult, especially when many init files are present.

Add `|| mysql_error` to every execution path in docker_process_init_files so that on failure the log clearly names the file that failed:

  [ERROR] [Entrypoint]: entrypoint.sh: failed while running /docker-entrypoint-initdb.d/02_seed.sql

Applies to: executable .sh, sourced .sh, .sql, .sql.gz, .sql.bz2,
            .sql.xz, and .sql.zst file types.

### DIFF
--- a/8.4/docker-entrypoint.sh
+++ b/8.4/docker-entrypoint.sh
@@ -66,17 +66,17 @@ docker_process_init_files() {
 				# https://github.com/docker-library/postgres/pull/452
 				if [ -x "$f" ]; then
 					mysql_note "$0: running $f"
-					"$f"
+					"$f" || mysql_error "$0: failed while running $f"
 				else
 					mysql_note "$0: sourcing $f"
-					. "$f"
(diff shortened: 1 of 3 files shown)


tianon:
Every one of these starts with `mysql_note "$0: running $f"` :sweat_smile:

(so if you get a failure, you can read the log backwards to see which one failed)

PR #1142, opened 2026-06-19
Adding support for database startup scripts

ikemtz:
There are instances where you want to run scripts when the container starts, and not just when the database is first created.  Here are some examples where this feature would be useful:

- You're starting an e2e (end-to-end) test run and you need to get the database to a "good" state.
- You're using an ORM (object-relational-mapper) library like [entity framework](https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/applying?tabs=dotnet-core-cli#idempotent-sql-scripts) that generates an idempotent that should be run on every startup to ensure the database is in a good state.

This merge request adds this capability by creating a new "/docker-entrypoint-startdb.d/" folder; files placed in this folder will be run on every container startup.

NOTE: The DockerHub image page description will need to be updated, with text that looks similar to this:

```txt
Running scripts on database startup
Each time this container image is started, it will execute files with extensions .sh, .sql, .sql.gz, .sql.bz2, .sql.xz, and .sql.zst that are found in /docker-entrypoint-startdb.d. Files will be executed in alphabetical order. When parsing .sh files without the execute bit set, they are sourced rather than executed.
```

### DIFF
--- a/8.4/Dockerfile.oracle
+++ b/8.4/Dockerfile.oracle
@@ -92,6 +92,7 @@ RUN set -eux; \
 	chmod 1777 /var/lib/mysql /var/run/mysqld; \
 	\
 	mkdir /docker-entrypoint-initdb.d; \
+	mkdir /docker-entrypoint-startdb.d; \
 	\
 	mysqld --version; \
 	mysql --version
--- a/8.4/docker-entrypoint.sh
+++ b/8.4/docker-entrypoint.sh
@@ -56,7 +56,7 @@ _is_sourced() {
 docker_process_init_files() {
 	# mysql here for backwards compatibility "${mysql[@]}"
 	mysql=( docker_process_sql )
-	
+
 	echo
 	local f
 	for f; do
@@ -286,7 +286,7 @@ docker_process_sql() {
 		set -- --database="$MYSQL_DATABASE" "$@"
 	fi
 
-	mysql --defaults-extra-file=<( _mysql_passfile "${passfileArgs[@]}") --protocol=socket -hlocalhost -uroot --socket="${SOCKET}" --comments "$@"
+	mysql --defaults-extra-file=<( _mysql_passfile "${passfileArgs[@]}") --protocol=socket -uroot -hlocalhost --socket="${SOCKET}" --comments "$@"
 }
 
 # Initializes database with timezone info and root password, plus optional extra db/user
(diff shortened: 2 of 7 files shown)


tianon:
This ("`always-initdb.d`" functionality) is intentionally left for downstream images to implement.  The root issue is that running scripts against a live database requires starting the server in a temporary local-only mode, doing the work, stopping it, then starting the real process -- and we're not willing to impose that overhead on every container startup for a feature most users don't need.

`docker-entrypoint.sh` is designed for exactly this kind of extension, though.  The `_is_sourced()` / `_main()` guard at the bottom of the script exists so you can `source` it, call its exported functions directly, and build your own startup sequence around them.  `docker_process_init_files` is the relevant public function -- its own docstring already uses `/always-initdb.d/*` as the example invocation:

```
# usage: docker_process_init_files [file [file [...]]]
#    ie: docker_process_init_files /always-initdb.d/*
```

A full custom entrypoint implementing this for MySQL (the equivalent for PostgreSQL is worked out in [docker-library/postgres#496](https://github.com/docker-library/postgres/pull/496), and the same `source` pattern ships as [`docker-ensure-initdb.sh`](https://github.com/docker-library/postgres/pull/1150) in that image):

```bash
#!/usr/bin/env bash
set -Eeuo pipefail

source /usr/local/bin/docker-entrypoint.sh

if [ "${1:0:1}" = '-' ]; then
	set -- mysqld "$@"
fi

if [ "$1" = 'mysqld' ] && ! _mysql_want_help "$@"; then
	mysql_check_config "$@"
	docker_setup_env "$@"
	docker_create_db_directories "$@"

	if [ "$(id -u)" = '0' ]; then
	fi

	if [ -z "$DATABASE_ALREADY_EXISTS" ]; then
		docker_verify_minimum_env
		ls /docker-entrypoint-initdb.d/ > /dev/null
		docker_init_database_dir "$@"
		docker_temp_server_start "$@"
		mysql_socket_fix
		docker_setup_db
		docker_process_init_files /docker-entrypoint-initdb.d/*
		mysql_expire_root_user
		docker_temp_server_stop
	else
		docker_temp_server_start "$@"
		mysql_socket_fix
		docker_process_init_files /always-initdb.d/*
		docker_temp_server_stop
	fi
fi

exec "$@"
```

(See also: [docker-library/postgres#267 (comment)](https://github.com/docker-library/postgres/pull/267#issuecomment-283495192), [docker-library/postgres#191 (comment)](https://github.com/docker-library/postgres/issues/191#issuecomment-387221757), [docker-library/postgres#821 (comment)](https://github.com/docker-library/postgres/issues/821#issuecomment-788223006), [docker-library/postgres#929](https://github.com/docker-library/postgres/issues/929), [docker-library/postgres#955 (comment)](https://github.com/docker-library/postgres/issues/955#issuecomment-1111410293))

